PR #3, opened 2020-06-11
Merge PRs with the opam-repository master branch

talex5:
It now uses the upstream OCurrent, since the merging is now done by the CI, not in OCurrent. When `master` updates, the rebuild is done in the background (it still shows as successful in GitHub until the new result is known, rather than switching to pending).

The `sed` invocation didn't work for me, so I replaced it with some OCaml code, which is probably neater anyway. I added a check that the opam file is version 2 and that it's in the right place.

I updated the instructions for reproducing the build locally.

With these changes, I was able to run `opam-repo-ci-local` on a local clone of opam-repository.

Note: this also removes the `~enable_submodules` option. We might want to bring that back at some point, but I suspect that people trying to add submodules to opam-repository is not an urgent problem at the moment.

We may want to upstream other changes from the branch at some point too. In particular, the graph looks a bit messy. But let's get this working first, then update to the new solvers, then see if we still need that...

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -1,13 +1,19 @@
 open Lwt.Infix
 open Current.Syntax
 
-type key = {
-  src : Current_git.Commit.t;
-}
-
 let pool = Current.Pool.create ~label:"analyse" 2
 
+let ( / ) = Filename.concat
 let ( >>!= ) = Lwt_result.bind
+
+let read_file ~max_len path =
+  let ch = open_in path in
+  Fun.protect ~finally:(fun () -> close_in ch)
+    (fun () ->
+       let len = in_channel_length ch in
+       if len <= max_len then really_input_string ch len
+       else Fmt.failwith "File %S too big (%d bytes)" path len
+    )
 
 module Analysis = struct
   type t = {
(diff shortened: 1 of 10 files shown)


lib/build.ml
```diff
@@ -91,15 +96,22 @@ module Op = struct
...
                  docker build .@.@."
          Current_git.Commit_id.pp_user_clone (Current_git.Commit.id commit)
+         master
          Dockerfile.pp (make_dockerfile ~for_user:true));
     let dockerfile = Dockerfile.string_of_t (make_dockerfile ~for_user:false) in
     Current.Job.start ~timeout:build_timeout ~pool job ~level:Current.Level.Average >>= fun () ->
-    Current_git.with_checkout ~enable_submodules:false ~job commit @@ fun dir ->
+    Current_git.with_checkout ~job commit @@ fun dir ->
+    (* Merge with master. The merge should work because we already tried this in the analysis step. *)
+    let cmd = "", [| "git"; "merge"; master |] in
```
kit-ty-kate, line 112: does this really work? I remember trying something like this but it doesn't work if the branch tested is older than the master branch as `Current_git.with_checkout` only fetches one branch at a time. Has this changed or was I missing something when I tried this back then?

lib/analyse.ml
```diff
@@ -27,51 +33,102 @@ module Analysis = struct
...
+        |> List.sort_uniq OpamPackage.compare
+        |> List.map (fun pkg ->
+            let path =
+              Printf.sprintf "packages/%s/%s/opam"
+                (OpamPackage.name_to_string pkg)
+                (OpamPackage.to_string pkg)
+            in
+            (* Check it exists, parses, and is the right version. *)
+            let content = read_file ~max_len:102400 (Fpath.to_string dir / path) in
+            let opam = OpamFile.OPAM.read_from_string content in
```
kit-ty-kate, line 79: I think there should be a filter in case the file does not exist then it should be removed from the list (using `git diff --diff-filter d` is not enough because we want to check packages if only something in its `files` directory changed as well).

Also if the directory exists but the opam file does not this should be an error

lib/analyse.ml
```diff
@@ -1,14 +1,20 @@
...
 
+let ( / ) = Filename.concat
 let ( >>!= ) = Lwt_result.bind
 
+let read_file ~max_len path =
+  let ch = open_in path in
+  Fun.protect ~finally:(fun () -> close_in ch)
+    (fun () ->
+       let len = in_channel_length ch in
+       if len <= max_len then really_input_string ch len
```
kit-ty-kate, line 14: shouldn't we use Lwt functions instead?

lib/build.ml
kit-ty-kate: Oh, I would also recommend to use the `-q` argument to the git merge

lib/analyse.ml
```diff
@@ -27,51 +33,102 @@ module Analysis = struct
...
-    let opam_files =
-      String.split_on_char '\n' output
-      |> List.filter (fun s -> not (String.equal s ""))
-      |> List.sort String.compare
-    in
-    let r = { opam_files } in
-    Current.Job.log job "@[<v2>Results:@,%a@]" Yojson.Safe.(pretty_print ~std:true) (to_yojson r);
-    Lwt.return (Ok r)
+    let master = Current_git.Commit.hash master in
+    let cmd = "", [| "git"; "merge"; master |] in
```
kit-ty-kate, line 48: Some pair of comment as in https://github.com/ocurrent/opam-repo-ci/pull/3/files#r438897884

lib/build.ml
talex5: Good point. My test-case is indeed simply ahead of master. I guess a `git fetch origin master` first would fix that (pulling from the local clone).

talex5: OK, I tested this and it does work as is. You have to use `git clone --single-branch` to disable that behaviour. The `master` hash comes from a separate fetch operation, so we know the repository we're cloning from already has it.

I've added `-q` as you suggested.

lib/analyse.ml
talex5: Could do. Feel free to make a PR! I just copied this code from ocaml-ci, where it doesn't seem to be introducing any noticeable latency.

lib/analyse.ml
talex5: Done.

PR #5, opened 2020-06-12
Use lwt.unix instead of Stdlib to handle IO

kit-ty-kate:
Some PR might change or add hundreds of opam files in a regular basis in opam-repository. The following PR fixes some latency issue that might arise in those cases.

Similar to https://github.com/ocurrent/ocaml-ci/pull/207, discussed in https://github.com/ocurrent/opam-repo-ci/pull/3#discussion_r438903420

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -7,12 +7,16 @@ let ( / ) = Filename.concat
 let ( >>!= ) = Lwt_result.bind
 
 let read_file ~max_len path =
-  let ch = open_in path in
-  Fun.protect ~finally:(fun () -> close_in ch)
-    (fun () ->
-       let len = in_channel_length ch in
-       if len <= max_len then really_input_string ch len
-       else Fmt.failwith "File %S too big (%d bytes)" path len
+  Lwt_io.with_file ~mode:Lwt_io.input path
+    (fun ch ->
+       Lwt_io.length ch >>= fun len ->
+       let len =
+         if len <= Int64.of_int max_len then Int64.to_int len
+         else Fmt.failwith "File %S too big (%Ld bytes)" path len
+       in
+       let buf = Bytes.create len in
+       Lwt_io.read_into_exactly ch buf 0 len >|= fun () ->
+       Bytes.to_string buf
     )
 
 module OpamPackage = struct
@@ -40,7 +44,7 @@ module Analysis = struct
   let is_duniverse _ = false
 
   let ocamlformat_source _ = None
-  
+
   let check_opam_version =
     let version_2 = OpamVersion.of_string "2" in
     fun name opam ->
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Side note: the "Hide whitespace changes" github feature makes it easier to see what has actually changed.

lib/analyse.ml
```diff
@@ -7,12 +7,11 @@ let ( / ) = Filename.concat
...
-  let ch = open_in path in
-  Fun.protect ~finally:(fun () -> close_in ch)
-    (fun () ->
-       let len = in_channel_length ch in
-       if len <= max_len then really_input_string ch len
-       else Fmt.failwith "File %S too big (%d bytes)" path len
+  Lwt_io.with_file ~mode:Lwt_io.input path
+    (fun ch ->
+       Lwt_io.length ch >>= fun len ->
+       if len <= Int64.of_int max_len then Lwt_io.read ~count:max_len ch
```
talex5, line 13: Rather surprisingly, `Lwt_io.read ~count` does not read `count` bytes. You need to use `Lwt_io.read_into_exactly` instead.

> read ~count ic reads at most ~count characters from ic

However, if you don't specify `count` then it *does* read all the bytes!

kit-ty-kate: Does it really matter if it doesn't? I believe the semantics is "read as much as possible to a maximum length of `count`" and we already checked its length just before. Unless the file changes between the two calls I believe this should return a string exactly of size `len`.

talex5: It does matter. I already had this bug in the solver code :-( If the OS is busy, it can unpredictably return a short read.

kit-ty-kate: My bad. Given that `Lwt_io` is supposed to be somewhat higher level than `Lwt_unix`, I thought this function would read repetitively until the end of the file or `count`, not just read once.

Fixed now.

talex5: @aantron : the `Lwt_io.read` API has confused both of us (see above). Any ideas how this can be made safer?

aantron: I suppose we could quietly turn `read ~count` into `read_into_exactly` for strings. This would be still be consistent with the docs, which describe the behavior you observed. It looks like `read`'s two behaviors are really attempts to expose the implementation detail that with `~count`, `read` tries to do exactly one read, and without it, `read` reads in a loop. This is the cause of the confusion, and it was probably a mistake to overload `read` to have these two different behaviors. We should just make `read` always read in a loop, just with a limit or not.

Note that `read` can still return less than `count` bytes in case end of input is reached, which is probably the most common reason why this would happen. That wasn't what was biting you in this issue, was it?

talex5: In the bug I had, it was reading less than `count` when the system was under load IIRC. Perhaps the argument could be called `limit` when it's just a bound? e.g.

- `read_string : ?count:int -> t -> string` (read exactly `count` if given, or the whole file if not)
- `read_single : limit:int -> t -> string`

Though I'm not sure if we actually need the second one. I bet most people who want a single read are using `read_into` already. Not sure about the name either. `read_exactly` is clear when `count` is provided, but not for reading the whole file. Could be two functions I suppose.

In any case, it would be good to clarify in the docs that end-of-file isn't the only reason why you may get a short read.

aantron: I updated the docs for now in https://github.com/ocsigen/lwt/commit/2daa73b2aec332c0c929b73161fb261a8c798e6e. If this bites someone again, I'd probably choose to make `read ~count` read in a loop unless there is end of input or an error.

PR #9, opened 2020-06-15
Convert from GitHub app to OAuth

talex5:
Apps are easy for users to install on multiple organisations, but for opam-repo-ci we just want to monitor one repository. The new `--repository` option can be used to control which one.

### DIFF
--- a/service/main.ml
+++ b/service/main.ml
@@ -40,8 +40,9 @@ let has_role user = function
            ) -> true
     | _ -> false
 
-let main config mode app capnp_address github_auth =
-  let engine = Current.Engine.create ~config (Pipeline.v ~app) in
+let main config mode github repo capnp_address github_auth =
+  let repo = (github, repo) in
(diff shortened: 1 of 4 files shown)


avsm:
I thought that we require GitHub App permissions in order to use certain features such as GitHub Checks though -- https://developer.github.com/v3/checks/runs/

> Write permission for the Checks API is only available to GitHub Apps. OAuth Apps and authenticated users can view check runs and check suites, but they are not able to create them. If you aren't building a GitHub App, you might be interested in the Statuses API.

I can see how an oAuth option will make deployment easier for single-repo CIs, but I just want to make sure that we don't run into a roadblock in the future with integrating the (very useful looking) Checks API to report results.  This seems essential for (e.g.) integrating Camelus with ocaml-repo-ci as code review is only available via Checks.

talex5:
OK, let's close this then.

PR #16, opened 2020-10-27
Do not fail when requested package is not available on (restriction of opam list --installable --with-test)

kit-ty-kate:
`opam list --depends-on <pkg> --installable --with-test` can give packages that are not available with the tests but that are installable without. So when we require repdeps to install without tests it should not fail but with it might.


talex5:
Needs rebasing (sorry - I just merged the OCluster branch!)

kit-ty-kate:
I was just doing it ^^ Done.

PR #18, opened 2020-10-27
Don't test revdeps if main build fails

talex5:


service/pipeline.ml
```diff
@@ -130,7 +131,7 @@ let build_with_cluster ~ocluster ~analysis ~master source =
         and+ build = Node.of_job `Built image ~label:"build"
         and+ tests = Node.of_job `Built tests ~label:"tests"
```
kit-ty-kate, line 132: the same thing is required for the tests I believe

kit-ty-kate, line 132: the same thing is required for the tests I believe

edit: oops sorry for the double post

talex5: Tests should be covered already by the `after:image` bit:
```ocaml
        let tests =
          let spec = test_spec ~platform pkg ~after:image in
          Build.v ocluster ~label:"test" ~base ~spec ~master source
```

kit-ty-kate: Oh I see! Sorry for the disturbance.

kit-ty-kate (approved):

PR #19, opened 2020-10-27
Precisely track revdeps by only allowing to run revdep tests when actually possible

kit-ty-kate:
This PR improves several things:
* to get the list of revdeps, opam-repo-ci is now using the `--coinstallable-with <pkg>` option from `opam list` to filter out the packages that won't be able to be installed. https://github.com/ocurrent/opam-repo-ci/pull/16 previously fixes this issue by ignoring the packages. However this is inefficient but with this method we can get the list reliably. Shout-out to @rjbou and @AltGr who took some time to answer my question.
* it also splits the call to opam list into two:
  - the first one is to get the regular revdeps
  - and the second one is to get the packages that are revdeps and that can also be tested

This allows to know precisely which packages are compatible and for what. With this we can remove the ` || test $? -eq 20` magic added in #16 and be more efficient by not spawning new jobs for nothing.

I hope this doesn't bother the rewrite of the new analysis phase using opam-0install's solver too much. I'm guessing revdeps integration in the rewrite is gonna come a bit later so hopefully this is useful in the meantime.

### DIFF
--- a/lib/build.ml
+++ b/lib/build.ml
@@ -17,9 +17,13 @@ module Spec = struct
     with_tests:bool;
   } [@@deriving to_yojson]
 
+  type opam_list = {
+    list_with_tests : bool;
+  } [@@deriving to_yojson]
+
   type ty = [
     | `Opam_fmt of Analyse_ocamlformat.source option
-    | `Opam of [ `Build of opam_build | `List_revdeps ] * package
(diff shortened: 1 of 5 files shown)


talex5 (approved):
Looks good! Deploying now...

PR #21, opened 2020-10-28
Pipeline: Set the default OCaml compiler to 4.11 and add 4.12 (alpha)

kit-ty-kate:
Currently not many packages fail on 4.12 so it's safe to start testing now.


talex5:
Deployed and building now.

PR #22, opened 2020-10-29
Make all jobs low-priority for now

talex5:


talex5:
This is nearly ready to go live, so forget about this change.

PR #23, opened 2020-11-03
Only test revdeps on 4.11 to save the cluster

talex5:


talex5:
Merging this for now to help until we have better job priorities.

PR #25, opened 2020-11-03
Set revdeps jobs as non-urgent

kit-ty-kate:


talex5 (commented):
Putting this in value does mean it will have to rebuild all the jobs for the new urgency level.

I suggest either putting this in `t` (next to the timeout setting), or just calculating it from `spec` (i.e. use urgent=false if the build's revdep = None, otherwise use the priority).

Note that a job should only be urgent if it's not a revdep **and** it's a high-priority OCurrent job. Otherwise, rebuilds due to e.g. opam-repository merges will be forced to be urgent too.

avsm:
Needs a merge to resolve conflicts.

talex5:
This is less important now that we have fair queuing - opam-repo-ci jobs won't block other users of the cluster. So the only benefit would be to make the revdeps of one PR lower priority than the main build of a later PR. We should probably revisit this later if that turns out to be a problem.

kit-ty-kate:
Superseded by https://github.com/ocurrent/opam-repo-ci/pull/110

PR #26, opened 2020-11-04
Makes the user-facing logs prettier

kit-ty-kate:
Here how the web-ui for PRs currently look:
scrn-2020-11-04-17-21-51

With this change:
* `build (passing)` would merge with/become its parent (in the screenshot `sihl-email.0.1.7` and `sihl-queue.0.1.7`)
* the `list revdeps *` lines would disappear from the point of vue of the user.

This two change makes the web-ui should look much more user-friendly.

### DIFF
--- a/service/node.ml
+++ b/service/node.ml
@@ -1,14 +1,13 @@
 open Current.Syntax
 
+type action = {
+  job_id : Current.job_id option;
+  result : [`Built | `Checked] Current_term.Output.t;
+}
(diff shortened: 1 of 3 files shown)


talex5:
Seems to be some extra nesting:

ui

kit-ty-kate:
It's fix, sorry for that
scrn-2020-11-10-16-59-32

PR #28, opened 2020-11-11
Add a custom script to test local changes easily

kit-ty-kate:
Just in case this is worth having it inside the repo, otherwise we can always insert it in a wiki page or something

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ var
 *.cap
 .*.swp
 .vscode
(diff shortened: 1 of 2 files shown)


talex5:
The problem I see with this is it requires the GitHub app's private key, so it's only useful for us (and a bit risky even then).

I propose #29 as an alternative. That replaces the broken old opam-repo-ci-local command with a new one.

PR #33, opened 2020-11-18
Avoid unnecessary recompilation when pushing to master

kit-ty-kate:
Even if the recompilation were hidden like on `ocaml-ci`, in the case of `opam-repo-ci` it does not make much sense to do. Changes are sufficiently separated and trackable that it makes little sense to not just click the rebuild button instead.
Also none of the builds are cached so with large revdeps runs this will strain the cluster for nothing.

### DIFF
--- a/lib/build.ml
+++ b/lib/build.ml
@@ -80,7 +80,10 @@ let run_job ?buffer ~job build_job =
   | Ok _ as x -> Lwt.return x
 
 module Op = struct
-  type nonrec t = t
+  type nonrec t = {
+    config : t;
+    master : Current_git.Commit.t;
+  }
 
   let id = "ci-ocluster-build"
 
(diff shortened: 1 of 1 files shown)


talex5 (changes requested):
A better way to do this would be to move `master` to the `t` type. Putting it in the value but excluding it from the hash is likely to go wrong (for example, it might rebuild using the old master value because it thinks the cached version is the same).

Longer term, I think that perhaps what we really want here is just an ultra-low priority for background updates to `master`. That could be done with a slight modification to https://github.com/ocurrent/ocurrent/blob/master/lib/job.ml#L174 to allow a mode where it always waits for explicit confirmation before starting if `not t.urgent`. But it would also require a change to the web UI so you could see the previously-latched log in that case.

PR #35, opened 2020-11-18
Do not show non-started jobs

kit-ty-kate:
Tests show as "(not started)" when the main build failed. Same thing with the revdeps, they appear even before the main build finished, so if it fails they will also all fail. This is not what we want.

This PR should fix the issue, however I believe the jobs will still be started in the background which might bloat the cluster. @talex5 is that the case?

### DIFF
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -108,15 +108,26 @@ let combine_revdeps ~revdeps ~revdeps_with_tests =
   in
   Revdep.Map.bindings map
 
+let build_tests ~revdep ~ocluster ~master ~base ~platform ~pkg ~after source =
+  let tests =
+    let spec = test_spec ~platform ?revdep pkg ~after in
+    Build.v ocluster ~label:"test" ~base ~spec ~master source
+  in
+  let+ state = Current.state ~hidden:true after
+  and+ tests = Node.action `Built tests in
+  match state with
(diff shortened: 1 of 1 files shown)


talex5:
Why don't we want to show tests that can't start yet? It's useful to see what would happen once they do pass. Otherwise, it looks like your PR is nearly done, and then suddenly new tasks start appearing!

I think we should just remove the hyperlink from `tests (not started)` in the web UI, so it just renders as plain text (no point sending people to an error page).

kit-ty-kate:
> Why don't we want to show tests that can't start yet? It's useful to see what would happen once they do pass.

I don't personally think it is. There is too much clutter and it does not help to see what has not succeeded.

> Otherwise, it looks like your PR is nearly done, and then suddenly new tasks start appearing!

I think this is a non-issue. It already does that in the analysis phase and for revdeps, it will still show up as on-going in the build status on github anyway.

talex5:
I've added some summary information in #45, so this now only affects the "Full report" section.

Perhaps we should draw a distinction between "Not started yet" (waiting for main build) and "Will never start" (because the main build failed or that platform is skipped).

kit-ty-kate:
I feel like in terms of usage, #45 pretty much makes this PR obsolete, so I'll close it for now. Thanks a lot for that!

PR #37, opened 2020-11-18
Only list active accounts

talex5:
Just list the accounts being monitored at the current time.

In particular, this will stop the old `talex5` entry from appearing at https://opam.ci.ocaml.org/github.

This is a cherry-pick of https://github.com/ocurrent/ocaml-ci/commit/12f85b9ac3301ce5a4a807eba98aad20461284a1

### DIFF
--- a/lib/index.ml
+++ b/lib/index.ml
@@ -9,12 +9,10 @@ type t = {
   db : Sqlite3.db;
   record_job : Sqlite3.stmt;
   remove : Sqlite3.stmt;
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
I don't really understand the diff, but I have just one question: what would happen to the logs of non-active accounts? Do they get deleted?

talex5:
No, they don't get deleted. But you can't get to them any longer using the web UI.

PR #42, opened 2020-11-20
Add all the supported arches

kit-ty-kate:
Should add 32bit x86 and 32bit arm in the current state.

Supersedes https://github.com/ocurrent/opam-repo-ci/pull/41


talex5:
As this was merged, I'm assuming you want it live too, so I've pushed it to the `live` branch as well.

talex5:
Ah, I see. It doesn't work because the pool name is wrong. Will fix...

PR #45, opened 2020-11-23
Web UI: summarise results

talex5:
- If there are errors, show a quick summary of them at the top of the page.
- Otherwise, indicate whether there are tests still in progress or whether the tests have all passed.


kit-ty-kate (approved):
Ohhh, very very cool feature! It worked really well in https://github.com/ocaml/opam-repository/pull/17685

PR #46, opened 2020-11-23
Check revdeps recursively

kit-ty-kate:
I've seen quite a few issues recently after merging `opam-file-format.2.1.1`. This version broke many packages not relying on it specifically but rather later down the line (module name clash, undefined modules in some specific cases, …)

To fix this issue checking all the revdeps recursively seems like the way to go. This will put more strain to the cluster however, so we might want to monitor this a bit more closely.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -84,7 +84,7 @@ let revdeps ~with_tests ~base ~variant ~pkg =
       setup_repository ~variant
       @ [
         run "echo '@@@OUTPUT' && \
-             opam list -s --color=never --depends-on %s --coinstallable-with %s --installable --all-versions --depopts%s && \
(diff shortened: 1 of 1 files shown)


talex5:
Hmm, it seems that `--recursive` means that the list includes the package itself!

```
~ $ opam list --depends-on current_examples
# Packages matching: (installed | available) & depends-on(current_examples)
# No matches found
```
```
~ $ opam list --depends-on current_examples --recursive
# Packages matching: (installed | available) & rec-depends-on(current_examples)
# Name           # Installed # Synopsis
current_examples --          Example pipelines for OCurrent
```

kit-ty-kate:
arf, I'm on it. I've also opened an issue on opam's issue tracker.

PR #47, opened 2020-11-23
Give each revdep its own cache hint

talex5:
Otherwise we'll try to build them all on the same machine, which isn't useful for packages with thousands of revdeps.

### DIFF
--- a/lib/build.ml
+++ b/lib/build.ml
@@ -146,6 +146,7 @@ module Op = struct
     let cache_hint =
       let pkg =
         match ty with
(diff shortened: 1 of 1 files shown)


kit-ty-kate (approved):
Fair enough. It's a bit sad to have to rebuild the base package itself first on each machine but it's true this will speed up larger set of revdeps. Maybe we could improve this again later by only giving each revdeps its own cache hint if the number of revdeps for the particular package is larger than 15 or so for example?

talex5:
Yep. At the moment the cluster hard-codes the estimated times for cached and non-cached builds, based on ocaml-ci's patterns (where cached builds are massively faster). For revdeps, the win for caching is much smaller. I should really add the estimates to the job description instead (https://github.com/ocurrent/ocluster/issues/90).

PR #54, opened 2020-12-06
add support for installing some tests using opam 2.1

avsm:
The "extras" tests all now install using opam 2.1, as a smoketest for the beta branch.

Untested until I figure out local execution, hence a draft PR. (I won't have much time for the next few days, so feel free to take this over if anyone wants)

### DIFF
--- a/lib/build.ml
+++ b/lib/build.ml
@@ -14,7 +14,8 @@ module Spec = struct
 
   type opam_build = {
     revdep : package option;
-    with_tests:bool;
+    with_tests : bool;
+    upgrade_opam : bool [@default false];
   } [@@deriving to_yojson]
 
   type opam_list = {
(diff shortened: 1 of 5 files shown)


talex5:
I've rebased it and tested locally. I just moved the flag to `ty` to shrink the diff a bit.

talex5:
https://opam-repo-ci.ci3.ocamllabs.io/job/2020-12-07/122519-ci-ocluster-build-437270
```
  * Incompatible packages:
    - (invariant) -> ocaml-variants
    - (request)
    You can temporarily relax the switch invariant with `--update-invariant'
```

kit-ty-kate:
I'm not entirely convinced by this feature. I'm already testing all packages with opam 2.1 with opam-health-check so this extra check seems redundant.
One thing i would really really enjoy having would be to have opam 2.1 used by default on opam-repo-ci. This would remove most of the annoying failures happening because opam 2.0 is not clever enough (wrt. depexts)

talex5:
Hmm, but then we aren't testing packages with the version of opam that people will actually use to install them.

kit-ty-kate:
> Hmm, but then we aren't testing packages with the version of opam that people will actually use to install them.

we could do the opposite of what this PR does and make opam 2.0 an extra check.

avsm:
Let's not drop opam 2.0 testing for quite some time please.  I'd like opam 2.1 to go in gradually, and _alongside_ the existing opam 2.0 testing.  If there's a failure, then it's something our users experience too.

avsm:
Changes look good to me, thanks @talex5.  I'm happy for this to be merged now so we have at least one test using opam 2.1beta, and for discussion about dropping opam 2.0 to happen at a later date.

PR #64, opened 2021-02-07
Dump the CUDF files on solver timeouts

kit-ty-kate:
This helps opam devs understand when they appear

cc @samoht


kit-ty-kate:
Pushed live

PR #66, opened 2021-02-08
Fallback to opam 2.1 when encountering solver timeouts

kit-ty-kate:
Temporary alternative to https://github.com/ocurrent/opam-repo-ci/pull/65
Proposal by @dra27

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -26,7 +26,6 @@ let opam_install ~variant ~upgrade_opam ~pin ~with_tests ~pkg =
           opam remove -y %s && opam install -y%s %s
           res=$?
         fi
-        test "$res" = 60 && (ls -lh /tmp/cudf-dump-*.cudf; cat /tmp/cudf-dump-1.cudf; exit 1)
         test "$res" != 31 && exit 1
         export OPAMCLI=2.0
         build_dir=$(opam var prefix)/.opam-switch/build
@@ -52,7 +51,6 @@ let setup_repository ~upgrade_opam =
   env "OPAMERRLOGLEN" "0" :: (* Show the whole log if it fails *)
   env "OPAMSOLVERTIMEOUT" "500" :: (* Increase timeout. Poor mccs is doing its best *)
   env "OPAMPRECISETRACKING" "1" :: (* Mitigate https://github.com/ocaml/opam/issues/3997 *)
-  env "OPAMCUDFFILE" "/tmp/cudf-dump" :: (* write CUDF files in case the solver timeouts *)
   [
     user ~uid:1000 ~gid:1000;
     copy ["."] ~dst:"/src/";
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -23,7 +23,7 @@ let opam_install ~variant ~upgrade_opam ~pin ~with_tests ~pkg =
         test "$res" = 0 && exit 0
         if test "$res" = 60 && diff -q /usr/bin/opam /usr/bin/opam-2.0; then
           sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam
-          opam remove -y %s && opam install -y%s %s
+          opam remove -y %s && opam install -y%s %s%s
           exit 1
         fi
         test "$res" != 31 && exit 1
@@ -38,7 +38,7 @@ let opam_install ~variant ~upgrade_opam ~pin ~with_tests ~pkg =
         exit 1
       |}
       pkg (if upgrade_opam then "install -y" else "depext -uivy") (if with_tests then "t" else "") pkg
-      pkg (if with_tests then "t" else "") pkg
+      pkg (if with_tests then "t" else "") pkg (if with_tests then "" else " && opam reinstall -yt "^pkg)
       (Variant.distribution variant)
   ]
 
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -82,17 +82,18 @@ let spec ~upgrade_opam ~base ~variant ~revdep ~with_tests ~pkg =
       @ tests
   }
 
-let revdeps ~with_tests ~base ~variant:_ ~pkg =
+let revdeps ~base ~variant:_ ~pkg =
   let open Obuilder_spec in
   let pkg = Filename.quote (OpamPackage.to_string pkg) in
-  let with_tests = if with_tests then " --with-test" else "" in
   { from = base;
     ops =
       setup_repository ~upgrade_opam:false
       @ [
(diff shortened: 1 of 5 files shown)


lib/opam_build.ml
```diff
@@ -21,6 +21,11 @@ let opam_install ~variant ~upgrade_opam ~pin ~with_tests ~pkg =
         opam remove -y %s && opam %s%s %s
         res=$?
         test "$res" = 0 && exit 0
+        if test "$res" = 60 && diff -q /usr/bin/opam /usr/bin/opam-2.0; then
+          sudo cp /usr/bin/opam-2.1 /usr/bin/opam
```
avsm, line 25: I believe this should be a hard link and not a cp (as that's what the base dockerfiles do)

dra27:
Just to check I've understood the script correctly: if opam 2.0 fails with a solver timeout, but opam 2.1 succeeds, the CI step will still fail with `exit 1`, yes? (I'm hoping that's true: i.e. the build still fails, but we can extract two pieces of info rather than one)

lib/opam_build.ml
```diff
@@ -21,6 +21,11 @@ let opam_install ~variant ~upgrade_opam ~pin ~with_tests ~pkg =
         opam remove -y %s && opam %s%s %s
         res=$?
         test "$res" = 0 && exit 0
+        if test "$res" = 60 && diff -q /usr/bin/opam /usr/bin/opam-2.0; then
```
dra27, line 24:
```suggestion
        if test "$res" != 0 && diff -q /usr/bin/opam /usr/bin/opam-2.0; then
```
?? Can this also be hit with depext-related errors too?

lib/opam_build.ml
kit-ty-kate: Fixed.

lib/opam_build.ml
kit-ty-kate: Testing seems to indicates that `opam-depext` also returns exit code 60 properly: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/597b4467d6799043fc07b4eafee2d7a038f95821/variant/compilers,4.04,mirage-conduit.3.0.0

kit-ty-kate:
> Just to check I've understood the script correctly: if opam 2.0 fails with a solver timeout, but opam 2.1 succeeds, the CI step will still fail with `exit 1`, yes? (I'm hoping that's true: i.e. the build still fails, but we can extract two pieces of info rather than one)

No it will ignore the error and just go with whatever opam 2.1 ends up doing.
The rational is that the solver timeout failures won't be fixed in opam 2.0 and above all we need testing the underlying package, not "knowing that this package makes the solver fail in CI". Solver timeouts can't be fixed (or very very very rarely) in the packages themselves so there is nothing we can do about it.
Continuing displaying these errors are mostly useless because they are so common that as opam maintainers we end up just ignoring them and there are so many of them on big runs that is makes detecting the real genuine failures more of a pain than it should.

lib/opam_build.ml
dra27: That wasn't quite what I meant - is solver timeout the only issue which can occur? I thought there were depext-related problems?

dra27:
- opam 2.0 has a solver timeout, so `$res` at L24 is 60 and you trigger your new branch
- opam 2.1 succeeds so `$res` will be `0` at L27
- L30 then runs `test "$res" != 31 && exit 1` and so fails the step

Did I miss something?

dra27:
If the fallback is to run opam 2.1 and ignore the fact that 2.0 failed, then this largely becomes a switch to opam 2.1 "by the backdoor".

> The rational is that the solver timeout failures won't be fixed in opam 2.0

Knowing that opam 2.0 is failing a lot of PRs increases the chance of the fixes (CUDF trimming, etc.) being back-ported!

I wasn't proposing that you start getting green ticks for solver timeouts, rather that if there's another error further down the line which 2.1 can reveal that we can have that instead.

lib/opam_build.ml
kit-ty-kate: ah, you're correct. Though, I would say I'd prefer to take one problem at a time. The depext issue is more difficult to detect and isolate.

kit-ty-kate:
>     * opam 2.0 has a solver timeout, so `$res` at L24 is 60 and you trigger your new branch
> 
>     * opam 2.1 succeeds so `$res` will be `0` at L27
> 
>     * L30 then runs `test "$res" != 31 && exit 1` and so fails the step
> 
> 
> Did I miss something?



>     * opam 2.0 has a solver timeout, so `$res` at L24 is 60 and you trigger your new branch
> 
>     * opam 2.1 succeeds so `$res` will be `0` at L27
> 
>     * L30 then runs `test "$res" != 31 && exit 1` and so fails the step
> 
> 
> Did I miss something?

oops, sorry about that, my bad.

dra27:
Excuse the lag, I went head down on OCaml's CI yesterday!

Sorry to push back on this, but the behaviour of the script before was what I was suggesting - if opam 2.0 is timing out, then it cannot be the case that CI passes the PR. The problem at the moment is that the solver timeout means you don't know if there's an actual problem with the package/revdep (i.e. a build problem) because you actually get to the build.

In other words, for example, looking at https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/769d51a59bbb306dea0cccd468f8480c5167bea8 (PR#18107): you presently have:

- 4.04
  x mirage-conduit.3.0.0 (failed: Sorry, resolution of the request timed out.)
    + tests (not started)

I'm suggest that opam 2.1 can be used to allow, for example, the tests to be started and run, but there must at the end be _at least_ the solver timeout error. It might want enhancing to say "Sorry, **opam 2.0** timed out resolving the request" and it may also be the the "Summary of errors" section wants to have a separate group for solver timeouts, so that it's easier for maintainers to see the important errors, but the error _cannot_ disappear because users may well see it.

kit-ty-kate:
My original rational was:
- timeouts have been there at least for the past 3 years and back then the answer I got on this is "it's a hard problem to fix in opam"
- opam 2.1 worked and ultimately it's the thing that is going to end-up being used by users
- I haven't seen any complaints from users outside CI about timeouts

I didn't really think the fix could be backported to 2.0, however Louis seem to agree with you so I opened https://github.com/ocaml/opam/pull/4538

--------

Other than that I agree on your idea, however the annoying thing is that tests of will end up not being tested as we skip failures and it is difficult to make a distinction in the code between "failure" and "failure but pass to the next". Maybe I'm ignoring something in the code that would help me do that though, I'll have a look.

kit-ty-kate:
I think it looks good enough now. I'll merge it tomorrow.

PR #67, opened 2021-02-09
Only run CI on a package when it significantly changed

kit-ty-kate:
This should help the cluster and avoid using energy for nothing


kit-ty-kate:
Pushed to production. It's working fine from what I can see.

The only issue (which isn't a big deal) is that PRs which only modifies non-significant parts of packages will be shown a red cross on github. To fix that and bring some new features I'll make a lint phase which will run in any case. This could be the basis of the Camelus-like integration in opam-repo-ci.

Issue #76, opened 2021-02-26
TODO: Fix revdeps


kit-ty-kate:
`opam list --depends-on <pkg> --recursive --with-test` make it so that every test dependencies of every reverse dependencies are returned. However this is not the same behaviour as `opam install --with-test` and results in opam-repo-ci testing too many revdeps. See: https://github.com/ocaml/opam/issues/4543

TODO: The revdeps step should remove calls to `opam list --depends-on <pkg> --recursive ...` and replace it by roughly:
```
l = opam list --depends-on <pkg> --with-test # without recursive
foreach x in l:
  lrec += opam list --depends-on <x> --recursive
l + lrec
```

avsm:
This solution sounds right to me! It's also very useful to informing the opam 2.2+ evolution of the CLI, since we could have this calculation in a single CLI invocation in the future...

AltGr:
Note that for `--required-by foo`, having the `--with-test` flag behave similarly as for `opam install` makes direct sense (if you replace "packages mentionned on the command-line" by "packages member of the arguments of `--required-by`).

Although we can intuitively give a meaning to it, in the case of reverse dependencies the similarity is less obvious, though, because it's not the test dependencies of `foo` that you want to consider, but the test dependencies of the other packages; it's much less obvious to compute (in 1 step I mean).

Maybe a better description would be to keep the behaviour of `--with-test`, but change the meaning of `--rec` so that it is never affected by `--with-*`.

PR #77, opened 2021-02-26
Remove unrelated revdeps from the list of revdeps to check

kit-ty-kate:
I'm quite embarrassed, the solution was so simple from the start... Thanks to @rjbou for the help!

Fixes https://github.com/ocurrent/opam-repo-ci/issues/76


avsm:
hah, I totally missed that too! Thanks @kit-ty-kate and @rjbou! :-)

PR #85, opened 2021-03-22
Eliminate blank line at end of Dockerfile

dra27:
cf. https://github.com/ocurrent/obuilder/issues/65

The continuation to a blank line generates a warning from docker - easiest just to eliminate the problem here!


dra27:
I don't think the ocaml-ci failure is mine?

kit-ty-kate:
Thanks!

Issue #73, opened 2021-02-15
TODO: Automatically check the dune constraint matches the dune-project file


kit-ty-kate:
To add to the `(lint)` phase

PR #90, opened 2021-04-03
linter: check that the dune constraint matches the dune-project file

kit-ty-kate:
Fixes https://github.com/ocurrent/opam-repo-ci/issues/73


avsm:
Very useful feature! In terms of long term maintenance, shouldn't this be integrated into opam-dune-lint so that this CI can just shell out to it?

kit-ty-kate:
> In terms of long term maintenance, shouldn't this be integrated into opam-dune-lint so that this CI can just shell out to it?

Indeed, this would be a nice addition in the future (though we might want to think about the implications of requiring all the explicit dependencies to be listed, first; this might be a bit too verbose for some packages living in one repository for instance).

I've opened https://github.com/ocurrent/opam-dune-lint/issues/28 to keep track of this. Merging for now as it seems to be working well so far

PR #94, opened 2021-04-11
Enable the opam sandbox

kit-ty-kate:
obuilder supports bubblewrap as opposed to docker, so the technical reason why we were disabling the opam sandbox doesn't hold anymore in our particular context :tada:


talex5:
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/766a299dae4c604d7fd0e72a3236a01dd16a65ed/variant/distributions,alpine-3.12,extunix.0.3.1 says:

> bwrap: capset failed: Operation not permitted

Is that related to this change?

(from https://github.com/ocaml/opam-repository/pull/18483)

PR #98, opened 2021-04-22
Wrap long tasks using Lwt_preemptive to avoid 504s when checking big PRs

kit-ty-kate:
download and extract functions from opam don't use Lwt and thus blocks the whole process and makes the web-ui stall.
`OpamFileTools.lint ~check_upstream:true` also uses download.


kit-ty-kate:
looks like it's working perfectly

PR #99, opened 2021-04-26
Improve the lint message on unmatched dune constraint

kit-ty-kate:
(suggestion by @rdicosmo)


lib/lint.ml
```diff
@@ -287,7 +287,7 @@ module Lint = struct
           Fmt.str "Warning in %s: The package has a dune-project file but no explicit dependency on dune was found." pkg
       | BadDuneConstraint (dep, ver) ->
           Fmt.str "Error in %s: Your dune-project file indicates that this package requires at least dune %s \
-                   but the dependencies only requires dune >= %s."
+                   but the dependencies only requires dune >= %s. Please check which requirement is the right one, and fix the other."
```
talex5, line 290:
```suggestion
                   but your opam file only requires dune >= %s. Please check which requirement is the right one, and fix the other."
```

kit-ty-kate:
Thanks a lot for the improvements!

PR #100, opened 2021-04-26
Set a package as significantly changed if the depexts field changed

kit-ty-kate:


talex5 (approved):
Seems fine (once the CI failure is fixed).

Is this a bug in opam's `effectively_equal` though?

kit-ty-kate:
> Is this a bug in opam's `effectively_equal` though?

Per Louis' comment in DM `depexts` is a bit of a grey area, in opam 2.0 it was ok to ignore it when rebuilding, now it's less clear (or at least not in line with how `available:` is treated). In practice it's too often change in opam-repository to not be annoying to users for very little gain so `effectively_equal` is probably not going to be change without https://github.com/ocaml/opam/issues/4647 first (most likely in opam 2.2~alpha)

PR #106, opened 2021-05-13
Do not make the linter rebuild on every change in master

kit-ty-kate:
Rational: the number of merges in opam-repository is rather high and I always end up looking at logs and not seeing the linter failure because it is being rebuilt (btw is there any ways in ocurrent(_web?) to never show when a job is being rebuilt @talex5 ?)

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -248,9 +248,8 @@ module Lint = struct
       master : Current_git.Commit.t;
     }
 
-    let digest { master } =
+    let digest { master = _ } =
       let json = `Assoc [
-          "master", `String (Current_git.Commit.hash master);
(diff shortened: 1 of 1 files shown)


talex5 (changes requested):
I don't think this is a good way to do it. If you don't want to track this at all, it should be in the context (rather than keeping it in the value and then giving the wrong digest for it). That's how it's already done in `build.ml`.

kit-ty-kate:
Fair enough, sorry for that. Fixed in https://github.com/ocurrent/opam-repo-ci/commit/5d5ed9ea3b5520c613eb01c9fde2c6285a3fba4a

PR #110, opened 2021-05-26
Set the revdeps jobs as non-urgent on modified packages

kit-ty-kate:
Discussed at the last meeting as a replacement for #107 and #108.

This prioritizes truly important jobs: any jobs but the revdeps of modified packages


kit-ty-kate:
There are still some issues (revdeps running before other tasks, but I suspect the issue is deeper in the scheduler) but it's been running fine since so merging. We can always fix things later.

PR #111, opened 2021-06-01
Only enable the verbose mode for the requested package, not its dependencies

kit-ty-kate:
This should reduce the verbosity of builds a little bit. However it might make CI behaviour change slightly compared to a simple `opam install <pkg>` (see https://github.com/ocaml/opam/pull/4564 for instance)

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -37,7 +37,7 @@ let opam_install ~variant ~upgrade_opam ~pin ~lower_bounds ~with_tests ~pkg =
           fi;
         done;
         exit 1|}
-      pkg (if upgrade_opam then "install -vy" else "depext -ivy") (if with_tests then "t" else "") pkg
+      pkg (if upgrade_opam then "install -v" else "depext -ivy") (if with_tests then "t" else "") pkg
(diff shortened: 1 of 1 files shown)


talex5 (commented):
@avsm mentioned today that ideally we should use the same command to install that users will. With `OPAMERRLOGLEN` being set in the base images we will already see errors on failure.

On the other hand, that won't show warnings, which might be useful.

@kit-ty-kate : what do you think about just removing `-v` completely? Is the verbose output useful in addition to the error log on failure?

kit-ty-kate:
The issue with that is that it would only display the log for the failing command and not for whatever successful commands there is before. For instance it is annoying in the case of a buggy `./configure` script that wasn't run with `sh -e` and would successfully finish but would display hints of what's going on on the next command (that is failing, e.g. `make`)

talex5:
> For instance it is annoying in the case of a buggy ./configure script that wasn't run with sh -e and would successfully finish but would display hints of what's going on on the next command (that is failing, e.g. make)

Ah, I see. I assumed opam showed the log output for the whole package, not just one step.

PR #114, opened 2021-07-14
Start all the jobs at the same time

kit-ty-kate:
This avoids a long-standing issue in opam-repo-ci: waiting for one particular PR to be checked can take a long time if other PRs come in the mix in the middle (e.g. PR#1 is still testing the main compiler checks when big-PR#2 arrives and stalls any further checks such as lower-bounds, tests, …). This makes the PR checking process much more linear and avoids bottlenecks.


talex5:
Rebasing this PR should fix the CI problem.

PR #115, opened 2021-07-14
Avoid the jobs to be all rebuilt at the same time every week

kit-ty-kate:
`Docker.peek` in `service/pipeline.ml` go fetch all the base images on a weekly basis. The issue is that every week this makes all the jobs rerun, and worse the web-ui does not show the previously available failures (this is another issue but I'm unsure how to fix this so this is hitting two birds with one stone)


kit-ty-kate:
Tested successfully

PR #116, opened 2021-07-15
Add macOS support

kit-ty-kate:
This is just a draft PR to visualize the remaining differences and test on a local instance first.

cc @patricoferris

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -45,8 +45,8 @@ let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~pkg =
 let setup_repository ~variant ~for_docker ~opam_version =
   let open Obuilder_spec in
   let home_dir = match Variant.os variant with
-    | `macOS -> "/Users/mac705"
-    | `linux -> "/home/opam"
+    | `macOS -> None
+    | `linux -> Some "/home/opam"
   in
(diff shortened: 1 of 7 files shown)


kit-ty-kate:
Checklist for macOS CI:

Critical:
- [ ] Fix and debug the new brew issue

Urgent but non-essential:
- [ ] Upgrade the macos-infra scripts so that it creates and upload a docker image to `ocaml/opam` at the end
- [ ] Replace the zfs backend by a new rsync backend to avoid slowdowns and other issues

lib/opam_build.ml
```diff
@@ -44,7 +44,32 @@ let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~pkg =
 
 let setup_repository ~variant ~for_docker ~opam_version =
   let open Obuilder_spec in
+  let home_dir = match Variant.os variant with
+    | `macOS -> "/Users/mac705"
```
kit-ty-kate, line 48:
```suggestion
    | `macOS -> "/Users/mac705"
```
@patricoferris is that always the case?

kit-ty-kate:
Keeping this PR around until all the things are fixed is becoming too costly in maintenance time. All the issues have their TODO comments so I think it’s good enough to merge this now and fix things later

PR #121, opened 2021-08-11
Add support for s390x testing for packages.

tmcgilchrist:


kit-ty-kate:
i thought the capacity of the s390x machines were rather small. Do you have the numbers on hand?

talex5:
We've got 4 (fast) cores, so at least twice as much as we've got for mac builds.

kit-ty-kate:
the difference is that macos doesn't compete with the other services (ocaml-ci and docker-base-images). But I guess we'll try

kit-ty-kate:
This PR also does the following changes:
* Add OpenSUSE 15.3
* Remove Ubuntu 20.10

All in all, it looks good now. Let's test this. Thanks!

PR #129, opened 2021-08-13
Make the version of opam used explicit in prevision for the upgrade of the base docker images

kit-ty-kate:
This way we don't have to change the code for opam 2.2 or if the docker images upgrade to opam 2.1 by default.

However this changes the key of the ocurrent cache and will restart all the builds. **To merge when we have a low number of small PRs.**


kit-ty-kate:
Everything restarted already so might as well merge this

PR #143, opened 2021-12-09
Make the cluster distribution fairer

kit-ty-kate:
The previous cache hint wasn’t too good. It always chose one server on the cluster when building one package on a given distribution. For example this causes problems when adding more builders (e.g. macos) as none of the build get passed over to the new builder.

### DIFF
--- a/lib/build.ml
+++ b/lib/build.ml
@@ -151,7 +151,7 @@ module Op = struct
         | `Opam (`List_revdeps, pkg)
         | `Opam (`Build _, pkg) -> OpamPackage.to_string pkg
       in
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
@talex5 do we even need a cache hint at all?

talex5:
Apart from anything, it's useful to see the cache hint in the scheduler status to see which jobs are which!

As usual, fixing https://github.com/ocurrent/ocluster/issues/90 would probably make this problem go away, because opam-repo-ci could indicate that a warm cache isn't very useful to it.

kit-ty-kate:
Fair enough. Let’s merge that for now then

PR #148, opened 2022-01-17
Fix crashes in compilers heavy PRs

kit-ty-kate:
Just a hunch on why https://github.com/ocaml/opam-repository/pull/19855 might make the CI crash.
In any case this is an obvious improvement


kit-ty-kate:
That did fix it :tada:

PR #152, opened 2022-02-03
Add Ubuntu 22.04 + Fedora 35

kit-ty-kate:


kit-ty-kate:
@dra27 could you give some hint about how to use the new ocaml-dockerfile API?

kit-ty-kate:
I had a shot at fixing it and I find it rather confusing. I think most functions dealing with the new `t` type should have open polymorphic variants instead of closed ones

kit-ty-kate:
@dra27 Thanks a lot for the fix! :heart:

dra27:
No probs. Having had a little further dabble with `Dockerfile_distro`, because we ideally want to have, say, `` `Ubuntu `V2204`` in both types but just `` `Ubuntu `LTS`` in one, I don't _think_ that can ever be done without a coercion. I think it could be done without coercion if we had `` `Ubuntu_2204`` and then `` `Ubuntu_LTS`` but that's clearly inferior for other reasons!

PR #156, opened 2022-02-08
Build the list of packages to test with both opam 2.0 and 2.1

kit-ty-kate:
opam 2.1 is taking way too long to run opam list twice at the moment but 2.1.3 should be able to fix this issue.

This would fixes issues where e.g. mirage 4.0.0~beta3 is not tested in the revdeps because `opam list --depends-on` is only ran with opam 2.0

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -105,12 +105,11 @@ let spec ~for_docker ~opam_version ~base ~variant ~revdep ~lower_bounds ~with_te
     @ tests
   )
 
-let revdeps ~for_docker ~base ~variant ~pkg =
+let revdeps ~for_docker ~opam_version ~base ~variant ~pkg =
   let open Obuilder_spec in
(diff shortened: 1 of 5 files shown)


lib/opam_build.ml
```diff
@@ -105,12 +105,11 @@ let spec ~for_docker ~opam_version ~base ~variant ~revdep ~lower_bounds ~with_te
     @ tests
   )
 
-let revdeps ~for_docker ~base ~variant ~pkg =
+let revdeps ~for_docker ~opam_version ~base ~variant ~pkg =
   let open Obuilder_spec in
   let pkg = Filename.quote (OpamPackage.to_string pkg) in
   Obuilder_spec.stage ~from:base (
-    (* TODO: Switch to opam 2.1 when https://github.com/ocaml/opam/issues/4311 is fixed *)
```
kit-ty-kate, line 112: fixed but not yet released

PR #165, opened 2022-04-11
Add more missing matching log rules

kit-ty-kate:


service/main.ml
```diff
@@ -26,56 +26,68 @@ let has_role user = function
...
+        report = {|\1|};
+        score = 35;
+      };
+      { (* Generic errors caught by opam (e.g. gcc) *)
+        pattern = {|[\n]# .+: error: (.+)[\n]|};
+        report = {|\1|};
+        score = 30;
+      };
+      { (* Opam errors *)
+        pattern = {|[\n]\[ERROR\] (.+)[\n]|};
```
kit-ty-kate, line 84: for some reason this pattern seems to eliminate all other ones with `[ERROR]` even with a higher score..

service/main.ml
```diff
@@ -26,56 +26,73 @@ let has_role user = function
...
```
kit-ty-kate, line 89:
```suggestion
        pattern = {|[\n]\[ERROR\] (.+)\.[\n]|};
```

service/main.ml
```diff
@@ -26,56 +26,73 @@ let has_role user = function
...
+        score = 100;
+      };
+      { (* Ignore failures on failing packages when the platform being tested is contained in the x-ci-accept-failures optional field *)
+        (* See lib/opam_build.ml corresponding bash command printing this exact line *)
+        pattern = {|[\n]A package failed and has been disabled for CI using the 'x-ci-accept-failures' field\.[\n]|};
+        report = {|[SKIP] Failure ignored|};
+        score = 100;
+      };
+      { (* Opam error: checksum does not match *)
+        pattern = {|[\n]\[ERROR\] (Failed to get sources of .+: Bad checksum)[\n]|};
```
kit-ty-kate, line 54:
```suggestion
        pattern = {|[\n]\[ERROR\] (Failed to get sources of .+: .+)[\n]|};
```

service/main.ml
```diff
@@ -26,56 +26,73 @@ let has_role user = function
...
+        report = {|[SKIP] Package not available|};
+        score = 100;
+      };
+      { (* Ignore failures on failing packages when the platform being tested is contained in the x-ci-accept-failures optional field *)
+        (* See lib/opam_build.ml corresponding bash command printing this exact line *)
+        pattern = {|[\n]A package failed and has been disabled for CI using the 'x-ci-accept-failures' field\.[\n]|};
+        report = {|[SKIP] Failure ignored|};
+        score = 100;
+      };
+      { (* Opam error: checksum does not match *)
```
kit-ty-kate, line 53:
```suggestion
      { (* Generic opam fetching error *)
```

service/main.ml
talex5: Perhaps the others don't have the `\n` at the start? Once there is a match starting at the `\n` I think it will skip ahead to after the matched text looking for the next one.

kit-ty-kate: ah :/ Then I suppose there is a bug when there are several regexp matching containing `\n`.

All the patterns have `\n` at the beginning:
Screenshot 2022-04-19 at 11 58 30

PR #167, opened 2022-05-05
Add pin for ocaml-version git submodule

tmcgilchrist:
This should probably be pinned to the submodule rather than using the opam version of ocaml-version?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,9 @@ COPY --chown=opam \
 	ocurrent/current_slack.opam \
 	ocurrent/current_web.opam \
 	/src/ocurrent/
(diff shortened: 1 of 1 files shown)


dra27:
The others will need pinning too, or - cf. the pins added by ocaml-ci in https://ci.ocamllabs.io/github/ocurrent/opam-repo-ci/commit/2ebf7e7edef099992049f1ca734b178c01f38bd5/variant/debian-11-4.14.

Probably worth doing properly inasmuch as it'll be confusing the day one of ocluster, ocaml-version or ocaml-dockerfile adds a dependency which isn't also a dependency of either opam-repo-ci itself or ocurrent.

PR #169, opened 2022-05-12
Adds ability to cancel jobs associated to a commit.

novemberkilo:
Towards https://github.com/ocurrent/ocurrent/issues/286

See also https://github.com/ocurrent/ocaml-ci/pull/452

PR #170, opened 2022-06-13
Refactor after #169

kit-ty-kate:
@novemberkilo Thanks a lot for this work. This is very very much appreciated :heart:

That said I would have prefered to wait that i come back from my leave to merge it as I have a few comments and recommendations on the implementation. (even more as the changes are currently not in `live`/deployed and were not urgent)

If you have questions on my reformating/refactoring, feel free to ask.

I’ll give an overview on the main part of the refactoring in the code itself in a minute as well as ask a couple of questions

### DIFF
--- a/web-ui/github.ml
+++ b/web-ui/github.ml
@@ -340,7 +340,7 @@ let repo_handle ~meth ~owner ~name ~repo path =
     in
     let cancel_many (commit: Client.Commit.t) (job_infos : Client.job_info list) =
       Lwt_list.fold_left_s (fun (success, failed) job_info ->
-        can_cancel commit job_info >>= function
+        can_cancel commit job_info >>= function (* TODO: Can we replace these two calls to the scheduler with only one call to just cancel? *)
         | None -> Lwt.return (success, failed)
         | Some (ji, j) ->
             Current_rpc.Job.cancel j >|= function
--- a/web-ui/github.ml
+++ b/web-ui/github.ml
@@ -339,24 +339,19 @@ let repo_handle ~meth ~owner ~name ~repo path =
       | Aborted | Failed _ | Passed | Undefined _ -> Lwt.return None
     in
     let cancel_many (commit: Client.Commit.t) (job_infos : Client.job_info list) =
-      let open Lwt.Infix in
-      Lwt_list.filter_map_p (fun job_info -> can_cancel commit job_info) job_infos >>= fun l ->
-        Lwt_list.map_p (fun (ji, j) -> Current_rpc.Job.cancel j |> Lwt_result.map @@ fun () -> ji) l >>= fun x ->
-          let success = ref [] in
-          let failed = ref 0 in
-          let log_error r =
-          match r with
-          | Ok(ji) -> success := ji :: !success
-          | Error (`Capnp ex) ->
-            incr failed;
-            Log.err (fun f -> f "Error cancelling job: %a" Capnp_rpc.Error.pp ex)
-          in
-          let _ = List.map log_error x in
-          Lwt.return (!success, !failed)
+      Lwt_list.fold_left_s (fun (success, failed) job_info ->
+        can_cancel commit job_info >>= function
+        | None -> Lwt.return (success, failed)
+        | Some (ji, j) ->
+            Current_rpc.Job.cancel j >|= function
+            | Ok () -> (ji :: success, failed)
+            | Error (`Capnp ex) ->
+                Log.err (fun f -> f "Error cancelling job: %a" Capnp_rpc.Error.pp ex);
+                (success, succ failed)
+      ) ([], 0) job_infos
     in
     Capability.with_ref (Client.Repo.commit_of_hash repo hash) @@ fun commit ->
-    let refs = Client.Commit.refs commit in
-    refs >>!= fun refs ->
+    Client.Commit.refs commit >>!= fun refs ->
     Client.Commit.jobs commit >>!= fun jobs ->
       begin cancel_many commit jobs >>= fun (success, failed) ->
         let open Tyxml.Html in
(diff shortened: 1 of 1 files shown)


web-ui/github.ml
```diff
@@ -337,24 +337,19 @@ let repo_handle ~meth ~owner ~name ~repo path =
...
+      Lwt_list.fold_left_s (fun (success, failed) job_info ->
+        can_cancel commit job_info >>= function
+        | None -> Lwt.return (success, failed)
+        | Some (ji, j) ->
+            Current_rpc.Job.cancel j >|= function
+            | Ok () -> (ji :: success, failed)
+            | Error (`Capnp ex) ->
+                Log.err (fun f -> f "Error cancelling job: %a" Capnp_rpc.Error.pp ex);
+                (success, succ failed)
+      ) ([], 0) job_infos
```
kit-ty-kate, line 349: This part of the code was previously going through the list of jobs 3 times (Lwt_list.filter_map_p, Lwt_list.map_p, List.map) for not much. In opam-repo-ci this list of jobs can be extremely long (500_000 or more) so this part of the code should be really focused on optimizing such list operations.

Furthermore using the parallel operations (Lwt_list.map_p) might overwhelm the ocluster server with too many operations at once, so I changed it to a sequential one instead (Lwt_list.fold_left_s)

web-ui/github.ml
```diff
@@ -337,24 +337,19 @@ let repo_handle ~meth ~owner ~name ~repo path =
...
-          match r with
-          | Ok(ji) -> success := ji :: !success
-          | Error (`Capnp ex) ->
-            incr failed;
-            Log.err (fun f -> f "Error cancelling job: %a" Capnp_rpc.Error.pp ex)
-          in
-          let _ = List.map log_error x in
-          Lwt.return (!success, !failed)
+      Lwt_list.fold_left_s (fun (success, failed) job_info ->
+        can_cancel commit job_info >>= function
```
kit-ty-kate, line 341: @novemberkilo I’m wondering if it might be more better to just call `Current_rpc.Job.cancel`. In the worse case scenario (not really, but see bellow) it does not do anything, whereas in the current state two rpc calls are done (one to see if the job is cancellable, and the other to cancel it).

In practice: https://github.com/ocurrent/ocurrent/blob/e0f32037e1fff945b53fbba13489aa8d250115ad/lib/job.ml#L92, shows that the "wrong cancellation" is logged. It might be nice to have a separate function that actually ignores when jobs are already done. Such function could be added to ocurrent as `Current_rpc.Job.cancel_if_cancellable` or something of the sort.

web-ui/github.ml
```diff
@@ -337,24 +337,19 @@ let repo_handle ~meth ~owner ~name ~repo path =
       | Aborted | Failed _ | Passed | Undefined _ -> Lwt.return None
     in
     let cancel_many (commit: Client.Commit.t) (job_infos : Client.job_info list) =
-      let open Lwt.Infix in
-      Lwt_list.filter_map_p (fun job_info -> can_cancel commit job_info) job_infos >>= fun l ->
-        Lwt_list.map_p (fun (ji, j) -> Current_rpc.Job.cancel j |> Lwt_result.map @@ fun () -> ji) l >>= fun x ->
-          let success = ref [] in
-          let failed = ref 0 in
```
kit-ty-kate, line 344: Also note that using List.fold_left is usually a bit faster than references due to the GC handling references less efficiently.

novemberkilo:
@kit-ty-kate thanks for improving upon #169 

This was one of the first issues I picked up and I spent most of my time trying to get my head around how to use`Lwt`and `capnp` - apologies for the manky formatting and inefficient code. Your comments and questions have helped me continue to learn as I get spun up on OCaml and its idioms -- thank you for this.

web-ui/github.ml
```diff
@@ -337,24 +337,19 @@ let repo_handle ~meth ~owner ~name ~repo path =
...
+                Log.err (fun f -> f "Error cancelling job: %a" Capnp_rpc.Error.pp ex);
+                (success, succ failed)
+      ) ([], 0) job_infos
     in
     Capability.with_ref (Client.Repo.commit_of_hash repo hash) @@ fun commit ->
-    let refs = Client.Commit.refs commit in
-    refs >>!= fun refs ->
+    Client.Commit.refs commit >>!= fun refs ->
     Client.Commit.jobs commit >>!= fun jobs ->
       begin cancel_many commit jobs >>= fun (success, failed) ->
```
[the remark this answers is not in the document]
talex5: I suspect the original idea was to pipeline the `refs` and `jobs` calls e.g.
```ocaml
let refs = Client.Commit.refs commit in          (* Start getting refs *)
Client.Commit.jobs commit >>!= fun jobs ->       (* Get jobs *)
refs >>!= fun refs ->                            (* Finish getting refs *)
```
However, both the old and new code above don't do that, and waste a round-trip.

web-ui/github.ml
talex5: It's a shame there's no `Lwt_list.iter_n` to do batches, like there is for `Lwt_stream`.

web-ui/github.ml
novemberkilo: Thanks for the suggestion - I'll try this out on `ocaml-ci` as I bring your refactoring across to it.

web-ui/github.ml
novemberkilo: 🤦🏾‍♂️

web-ui/github.ml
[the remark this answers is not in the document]
novemberkilo: Ah this is just weird formatting left over. I pulled apart some things to try and understand them a bit better and this line didn't get returned to the state I found it in.

PR #172, opened 2022-06-17
Exclude dune itself from DuneConstraintMissing

emillon:
That is, it is ok for dune to have a dune-project file but not to depend on dune.

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -118,6 +118,9 @@ module Check = struct
         | exn -> Lwt.fail exn
         end
 
+  let is_dune name =
+    OpamPackage.Name.equal name (OpamPackage.Name.of_string "dune")
+
   let get_dune_constraint opam =
     let get_max = function
       | None, None -> None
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Sounds fair. Thanks a lot!

dra27:
I was just looking at the PR thinking "we should probably exclude Dune from that check" 🙂

lib/lint.ml
```diff
@@ -118,6 +118,9 @@ module Check = struct
         | exn -> Lwt.fail exn
         end
 
+  let is_dune name =
+    OpamPackage.Name.equal name (OpamPackage.Name.of_string "dune")
```
dra27, line 122: The compiler packages also trigger this one - cf. https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/6bac8b5b3330550680e20409e483074590e54b95/variant/(lint)

emillon: Since dune is an optional tool for the compiler, something like `depopts: ["dune" { with-tools }]` might make sense for it, but it's a can of worms.

PR #173, opened 2022-07-18
Updates

MisterDA:
- dune.3.3 (remove unused `db` field, remove empty `.mli`);
- ocurrent & ocluster:
  + new `Current_github.webhook` removes `~has_role` but requires exposing `Index.get_job_ids`;
  + new UI look-and-feel;
  + various changes.
- riscv64 definition, but disable jobs (the worker is too inefficient and would be overloaded).

### DIFF
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -232,7 +232,7 @@ let build_with_cluster ~ocluster ~analysis ~lint ~master source =
       ) (Ocaml_version.Opam.V2.switches `X86_64 default_compiler_full) @
       List.filter_map (function
         | `X86_64 -> None
-        | `Riscv64 -> None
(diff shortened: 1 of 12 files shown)


service/pipeline.ml
```diff
@@ -232,6 +232,7 @@ let build_with_cluster ~ocluster ~analysis ~lint ~master source =
       ) (Ocaml_version.Opam.V2.switches `X86_64 default_compiler_full) @
       List.filter_map (function
         | `X86_64 -> None
+        | `Riscv64 -> None
```
kit-ty-kate, line 235:
```suggestion
        | `Riscv64 -> None (* TODO: unlock this one when more machines are available *)
```

PR #179, opened 2022-08-05
Commit rebuild

kit-ty-kate:
Ports https://github.com/ocurrent/ocaml-ci/pull/469 (i hadn’t seen it before so i redid the commit instead of copying the code, sorry about that) as well as a few cleanup


kit-ty-kate:
cc @novemberkilo

PR #186, opened 2022-10-26
Vendor opam to bring the opam lint changes from opam 2.2.0

kit-ty-kate:


MisterDA:
You could add the branch name you want to track of the opam repo.

kit-ty-kate:
Is that what's bothering ocaml-ci?

PR #191, opened 2022-11-04
Do not run the tests if the package does not have any

kit-ty-kate:


lib/analyse.ml
```diff
@@ -173,6 +178,44 @@ module Analysis = struct
...
+  let map_has_tests ~dir (pkg, kind) =
+    let path = Fpath.to_string (package_to_path pkg) in
+    get_opam ~cwd:dir path >|= function
+    | Error () -> assert false
+    | Ok content ->
+        let filename = OpamFile.make (OpamFilename.raw path) in
+        let content = OpamFile.OPAM.read_from_string ~filename content in
+        let has_tests = has_tests content in
+        (pkg, {kind; has_tests})
+
```
kit-ty-kate, line 218: @rjbou could you have a look at this code to double-check there isn't something i forgotten? This code is suppose to check if a package has tests or not.
Is checking if the package contains `with-test` or has a non-empty `run-test` field enough? Is the `with-test` variable not substituted during parsing?

rjbou: You have full coverage, no `with-test` added, only resolved for filtering.
You don't use `OpamFileTools.all_variables` for optimisation purpose?

kit-ty-kate: Cool. `OpamFileTools.all_variables` is not exported so I can't use it here.

rjbou: oh ok

Issue #137, opened 2021-10-19
TODO: Test in a local switch inside a dune project


kit-ty-kate:
https://github.com/ocaml/opam-repository/issues/19809

cc @ejgallego

desirekaleba:
Hi @tmcgilchrist,

I am an Outreachy intern willing to contribute to `Improve OPAM Repository CI usability`. Can I be assigned to this issue, please?

patricoferris:
Hi @desirekaleba,

Do let us know if there's anything unclear about this issue. There's more information in the referenced opam-repository issue but it might not be enough. The first step is definitely getting things setup locally and hopefully the contributing guide can help with that https://github.com/ocurrent/opam-repo-ci/blob/master/CONTRIBUTING.md, beyond that if you have any doubts do ask on this issue :))

PR #200, opened 2023-01-03
Add a check using a local switch with spaces in the directory name

### LINKED ISSUE #199: Add spaces to opam root during CI
It is very common to have spaces in directory paths on Windows (ex. `C:\Users\Jane Smith`) and is not unusual on macOS either. Packages should not fail to build simply because a user has a space in their path.

The suggestion is to change the opam root to include a space. Examples:
* `/home/runner/work/has space`
* `/home/has space`
* `D:/a/has space`
* `C:/GitLab-Runner/builds/has space`

The suggestion is to make the default in a new version of CI (ex. `setup-ocaml.v2.1.0`) use a space since it is a bug to not support spaces. ~However, if the field `x-validation-path-with-spaces: false` is present in all `.opam` files (can use `awk` to check), then no space should be used.~

Cross-links:
* https://github.com/ocaml/setup-ocaml/issues/634
* https://github.com/diskuv/dkml-workflows/issues/8
* [OCaml Appveyor CI]( https://github.com/ocaml/ocaml/blob/a647c7fbec267080b2e891c2fc6e6ed91008f4d0/tools/ci/appveyor/appveyor_build.cmd#L24-L25) does something similar

kit-ty-kate:
Fixes #199 
Fixes https://github.com/ocurrent/opam-repo-ci/issues/137

### DIFF
--- a/lib/build.ml
+++ b/lib/build.ml
@@ -24,10 +24,12 @@ module Spec = struct
     revdep : package option;
     with_tests : bool;
     lower_bounds : bool;
+    local : bool;
     opam_version : [`V2_0 | `V2_1 | `Dev];
   } [@@deriving to_yojson]
 
   type list_revdeps = {
+    local : bool;
     opam_version : [`V2_0 | `V2_1 | `Dev];
   } [@@deriving to_yojson]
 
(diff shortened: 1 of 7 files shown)


kit-ty-kate:
cc @jonahbeckford

Requires https://github.com/ocaml/ocaml/pull/11590 to be backported and 4.14.2 & 5.0.1 to be released before this can be merged in the current state.

Et7f3:
This also requires https://github.com/ocaml/ocaml/pull/10727 that I need to split because the compiler installed doesn't works

dra27:
@Et7f3 - I think it should be fine, certainly for 5.0. ocaml/ocaml#11112 fixed the quoting of shebang lines. The fact 4.14 doesn't support it is less likely to be a problem - the compiler's build itself never calls shebang scripts, and virtually all tools will use the `.opt` versions of tools (we could also back-port the fix from 5.0 to 4.14 quite reasonably as a bug fix).

The only other change in https://github.com/ocaml/ocaml/pull/10727 that's not related to the testsuite is in `utils/ccomp.ml` and I'm not sure it's sound.

PR #205, opened 2023-01-30
Add local development setup script

art-w:
So, I'm cheap, I don't have an ocluster ready-to-go, or workers, or a nice filesystem for obuilder, and I keep forgetting all the commands... I've been putting together the following script to be able to setup everything and perhaps it'll be useful for others (or at least future me).

But I'm not saying that it works well! Just enough to quickly test things out. It's brittle, sometimes I have to `^C` and retry. There are probably other issues down the pipeline, but obuilder eats through my disk like peanut butter so I'm unable to really test that part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,13 @@ Status: **experimental**
 
 This is an [OCurrent][] pipeline that tests submissions to [opam-repository][].
 
+The analysis step will detect which packages have been changed.
+Then, for each supported platform it will try to install the package.
+If that succeeds, it will run the package's tests, and in parallel it will find other packages that
+depend on this one and test them too.
+
+## Manual local setup
+
 To test locally you will need:
 
 1. A [personal access token][] from GitHub.
@@ -25,11 +32,6 @@ Browse to http://localhost:8080 to see the web UI.
 You can either set the confirm threshold (at the bottom of the web page) to allow all builds to start,
 or just click on a yellow box and then on the `Start now` button to start one step manually.
 
-The analysis step will detect which packages have been changed.
-Then, for each supported platform it will try to install the package.
-If that succeeds, it will run the package's tests, and in parallel it will find other packages that
-depend on this one and test them too.
-
 ### Web UI
 
 The public web front-end is a separate process.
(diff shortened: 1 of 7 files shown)


service/main.ml
```diff
@@ -118,8 +118,11 @@ let main config mode app capnp_address github_auth submission_uri =
   add_default_matching_log_rules ();
   Lwt_main.run begin
     let listen_address = Capnp_rpc_unix.Network.Location.tcp ~host:"0.0.0.0" ~port:Conf.Capnp.internal_port in
-    Capnp_setup.run ~listen_address capnp_address >>= fun (vat, rpc_engine_resolver) ->
-    let ocluster = Capnp_rpc_unix.Vat.import_exn vat submission_uri in
+    Capnp_setup.run ~listen_address capnp_address >>= fun (_vat, rpc_engine_resolver) ->
+    let ocluster =
+      let vat = Capnp_rpc_unix.client_only_vat () in
+      Capnp_rpc_unix.Vat.import_exn vat submission_uri
```
art-w, line 124: :warning: I have no idea why it didn't work with the previous code for the `vat` and what impact this change will have on production, I need to check what's going on.

README.md
```diff
@@ -48,3 +48,40 @@ Then browse to http://localhost:8090/github to see the public UI.
 [OCurrent]: https://github.com/ocurrent/ocurrent
 [OCluster]: https://github.com/ocurrent/ocluster
 [opam-repository]: https://github.com/ocaml/opam-repository
+
+
+### Local development
```
talex5, line 53: It's a bit confusing to have a "Local development" section that differs from the local development setup described earlier in the file (which uses `opam-repo-ci-local` for a simpler but less complete approximation of the real system).

service/main.ml
art-w: With the previous code, I'm getting:
```
current_ocluster [WARNING] Error connecting to build cluster (will retry):
Failure("Failed: Unknown persistent service ID")
```

talex5: Are you using the same private key for both processes? If so, it will think you want it to connect to itself.

benmandrew:
This may be useful for future reference, but will not be merged as we now have a local development workflow.

PR #209, opened 2023-03-01
Always show the latest failure for the analysis step

kit-ty-kate:
Caused issue in https://github.com/ocaml/opam-repository/pull/23407/files#r1121707096 where the analysis failed but nothing was shown in the web-ui.

The idea for the fix is thus to keep the latched version of analysis to get the value out of it (so it doesn't remove the whole web-ui output if something went wrong during a rebuild of the analysis) and have the non-latched version to display in the web-ui so users know what went wrong.

Author: @art-w

### DIFF
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -225,8 +225,7 @@ let build_with_cluster ~ocluster ~analysis ~lint ~master source =
       ) acc [`Aarch64; `X86_64]
     ) [] default_compilers
   in
-  let analysis = Node.action `Analysed analysis
-  and lint = Node.action `Linted lint
+  let lint = Node.action `Linted lint
   and extras =
     let build ~opam_version ~distro ~arch ~compiler label =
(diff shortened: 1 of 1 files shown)


service/pipeline.ml
```diff
@@ -385,7 +383,11 @@ let test_pr ~ocluster ~master ~head =
     in
     Lint.check ~master ~packages src
   in
-  let builds = build_with_cluster ~ocluster ~analysis ~lint ~master commit_id in
+  let builds =
+    Node.root
+      (Node.leaf ~label:"(analysis)" (Node.action `Analysed latest_analysis)
+      :: build_with_cluster ~ocluster ~analysis ~lint ~master commit_id)
```
art-w, line 389: Perhaps the `lint` leaf could similarly be extracted from `build_with_cluster`? (or `latest_analysis` could be passed to that function to preserve the pretty `Node.root [...]`)

PR #210, opened 2023-03-03
Define `CI` and `OPAM_REPO_CI` env vars in runs

MisterDA:
Defining `CI` and `OPAM_REPO_CI` environment variables advertise to code running in CI jobs that they're running in CI. Test frameworks may use this to display all logs in the standard output/error instead of logging to files (that are inaccessible to the user), and display a maximum of information.


kit-ty-kate:
Thanks

PR #222, opened 2023-06-01
Update ocaml-dockerfile to include opensuse-tumbleweed. (Do-not-merge)

novemberkilo:
This update [breaks capnproto](https://github.com/ocaml/opam-repository/issues/23851) (and thus will likely break other packages for opensuse-tumbleweed). The root cause for this appears to be that the `os-family` has changed from `suse` to `opensuse` (see also https://github.com/ocaml/opam/issues/5565).


MisterDA:
There should be somewhere where you can filter out OpenSUSE Tumbleweed–and still update ocaml-dockerfile for the new distros definitions.

tmcgilchrist:
Fix for capnp in https://github.com/ocaml/opam-repository/pull/23852 and updates included in https://github.com/ocurrent/opam-repo-ci/pull/225

PR #230, opened 2023-08-07
Added FreeBSD

mtelvers:
Tested with `opam-repo-ci-local`.

### DIFF
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -280,7 +280,7 @@ let build_with_cluster ~ocluster ~analysis ~lint ~master source =
     Node.branch ~label:"compilers" compilers;
     Node.branch ~label:"distributions" linux_distributions;
     Node.branch ~label:"macos" macos;
-    Node.branch ~label:"FreeBSD" freebsd;
+    Node.branch ~label:"freebsd" freebsd;
     Node.branch ~label:"extras" extras;
   ]
 
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -136,6 +136,8 @@ let build_with_cluster ~ocluster ~analysis ~lint ~master source =
           match Variant.os variant with
           | `macOS ->
               Current.return (Build.MacOS (Variant.docker_tag variant))
+          | `FreeBSD ->
+              Current.return (Build.FreeBSD (Variant.docker_tag variant))
           | `linux -> (* TODO: Use docker images as base for both macOS and linux *)
               let+ repo_id =
                 Docker.peek ~schedule:weekly ~arch:(Ocaml_version.to_docker_arch arch)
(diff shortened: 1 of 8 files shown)


kit-ty-kate:
LGTM, pushed to live to see if it works.

service/pipeline.ml
```diff
@@ -264,6 +280,7 @@ let build_with_cluster ~ocluster ~analysis ~lint ~master source =
     Node.branch ~label:"compilers" compilers;
     Node.branch ~label:"distributions" linux_distributions;
     Node.branch ~label:"macos" macos;
+    Node.branch ~label:"FreeBSD" freebsd;
```
kit-ty-kate, line 283: Otherwise it becomes first in the list and that's not really what most people should see first.
```suggestion
    Node.branch ~label:"freebsd" freebsd;
```

kit-ty-kate (commented):
Could also add the `"(experimental)"` inscription that macos has in the web ui?

kit-ty-kate (approved):

kit-ty-kate:
Let's merge this tomorrow if the CI isn't caught in an infinite loop by then (looks good so far)

kit-ty-kate:
cc @hannesm we’re now testing FreeBSD in opam-repository

hannesm:
@kit-ty-kate great to hear. Congratulations.

hannesm:
It may be useful to record not only FreeBSD, but as well the major.minor thereof (IIUC this is 13.2 in your current setup). Also, I don't quite understand "ocurrent" in depth, but are there mechanisms in place to upgrade the system packages eventually / regularly (i.e. gmake / git / ..)? And the OS itself (`freebsd-update` is a great utility for binary updates)?

mtelvers:
The initial deployment was on 12.4; however, these have been upgraded to 13.2 this morning.

The machine(s) will be managed via Ansible scripts.  These are under [development](https://github.com/mtelvers/freebsd-infra.git).  They use `freebsd-update` to update the OS and build the base images.

tmcgilchrist:
https://github.com/ocurrent/obuilder/issues/109

hannesm:
Thanks for your work. I'm curious why FreeBSD is considered experimental, i.e. what is the process to be considered stable (is there anything I can help with)?

hannesm:
Thanks for your reply @kit-ty-kate. If you see strange FreeBSD issues in CI, feel free to tag me. But certainly a `availble: os != "freebsd"` is fine for some packages as well (if nobody bothers / it is too hard to fix).

tmcgilchrist:
Agreeing with the reasons that @kit-ty-kate has posted, 1-2 months while we work through the setup seems right.
We have 1 main server running FreeBSD 13.2 that is handling the load quite nicely but it is early days and we want to keep it out of the main loop for opam-repo. So marking it as experimental makes sense. There are some setup improvements we still need to make around the hardware to avoid killing as many NVME drives. 

Right now we support 4.14 and 5.0 on FreeBSD 13.2 x86. When 5.1 comes out will would swap 5.0 for that and continue with the 4.14 LTS and latest 5.* version. Do you think it is worth providing more OCaml versions? 

So far the performance of FreeBSD obuilder is roughly inline with Linux runc implementation, which is very nice. Unlike macOS and Windows 😞

PR #231, opened 2023-09-04
Use opam master on macOS and FreeBSD too

kit-ty-kate:
See https://github.com/ocurrent/opam-health-check/pull/60


mtelvers (approved):
LGTM - This logic is no longer required as macOS and FreeBSD have the development release of opam. 👍

PR #233, opened 2023-09-05
UI: Reorganize the summary of errors by type of errors to make things easier to read

kit-ty-kate:
Request by @UnixJunkie in https://github.com/ocaml/opam-repository/pull/24327#issuecomment-1705803328

Sadly this one makes things harder to read as the current error detection system isn't clever enough to regroup similar issues:
scrn-2023-09-05-16-03-43

### DIFF
--- a/web-ui/github.ml
+++ b/web-ui/github.ml
@@ -202,12 +202,22 @@ let link_jobs ~owner ~name ~hash ?selected jobs =
     in
     StatusTree.add k x trees
   in
+  let render_error trees { Client.variant; outcome } =
+    let k = [Fmt.str "%a" Client.State.pp outcome; variant] in
+    let x =
+      let uri = job_url ~owner ~name ~hash variant in
+      let label = String.split_on_char Common.status_sep variant in
+      let label = String.concat " / " label in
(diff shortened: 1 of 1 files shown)


benfaerber:
How do you think the errors should be grouped? Maybe it could a header per package and then group them by version beneath the header.

kit-ty-kate:
I think extending and improving the error detection system

https://github.com/ocurrent/opam-repo-ci/blob/980e10178b06f27d1b4161afd15afe0fab042082/service/main.ml#L28

would group most errors pretty well. Especially something to get either detect errors further or post-process the `The compilation of <pkg> failed at "..."` type errors would fix most errors i feel like.

Regrouping per package would definitely not be enough, you want to regroup by type of errors as it makes reading the output much easier.

UnixJunkie:
Yes, the important thing is regrouping by type of error.
Then you know how many errors there are to repair.

shonfeder:
Since this PR seems stalled, and more substantial design and  refactoring work is needed, I'm closing it in favor of https://github.com/ocurrent/opam-repo-ci/issues/312 for know. Please reopen if I'm mistaken and this can still move forward as is.

PR #237, opened 2023-09-13
Remove support for Oraclelinux

kit-ty-kate:
I'm not aware of anyone using it and too annoying to support as there no support for development packages.

Discussed in the opam-repository maintainers meeting (e.g. https://github.com/ocaml/opam-repository/pull/24360)
I think this is probably something we want to remove from ocaml-dockerfile as a whole, as the same failures will happen in OCaml-CI too.


emillon:
Hi.
This PR also removed support for Centos 7, does the same argument apply to it?

PR #242, opened 2023-09-19
Enable -warn-error +partial-match by default to ensure changes in a dependency do not lead into unforseen runtime failures

kit-ty-kate:
See https://github.com/ocaml/ocaml/issues/12475
Having this would've prevented https://github.com/ocaml/opam-repository/pull/24467 from occurring and warned the mdx maintainers that their package is not compatible yet.


shonfeder (approved):
Sounds like a good idea to me!

kit-ty-kate:
Requires https://github.com/ocaml/dune/pull/10489 (also a way to disable this feature for dune < "whatever version this is going to be fixed in")

shonfeder:
I see. Converted this to a draft in that case. Thanks.

PR #251, opened 2023-11-27
Decrement PR status metric when PR is merged

benmandrew:
Corrects https://github.com/ocurrent/opam-repo-ci/pull/247, which did not decrement the status counters when a PR was merged and removed from the active refs list. This led to old PR statuses being kept track of in the metrics.

### DIFF
--- a/lib/index.ml
+++ b/lib/index.ml
@@ -241,7 +241,7 @@ let set_active_refs ~repo (refs : (string * string) list) =
     (* Set difference: find refs that have been removed by merging *)
     let removed_refs =
       List.filter
-        (fun k -> not @@ List.exists ((=) k) refs)
+        (fun k -> not @@ List.mem k refs)
         old_refs
     in
     List.iter (fun (_, hash) ->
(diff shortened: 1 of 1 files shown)


lib/index.ml
```diff
@@ -225,9 +232,23 @@ let active_accounts = ref Account_set.empty
...
+let active_refs : (string * string) list Repo_map.t ref = ref Repo_map.empty
 
 let set_active_refs ~repo (refs : (string * string) list) =
+  (* Remove statuses from Status_cache corresponding to removed refs *)
+  Repo_map.find_opt repo !active_refs
+  |> Option.iter (fun old_refs ->
+    (* Set difference: find refs that have been removed by merging *)
+    let removed_refs =
+      List.filter
+        (fun k -> not @@ List.exists ((=) k) refs)
```
punchagan, line 244: nitpick: This could potentially also become 

```ocaml
(fun k -> not @@ List.mem k refs)
```

punchagan (approved):
Good catch, and the fix looks good to me! (apart from the minor nitpick). 

If you do end-up making the changes, I think it would be worth squashing all 3 commits into a single one, since these minor tweaks may not be useful history to keep.

lib/index.ml
benmandrew: Great, thanks! I was sure that function existed under some name or another...

Issue #260, opened 2024-01-22
Opam-repo-ci does not run on MacOS


benmandrew:
Disclaimer: this issue is with `opam-repo-ci` *itself* running on macOS, not with macOS jobs.

On macOS while running the lint job the service runs into an issue, which means that the process cannot be ctrl-c'd as well as the OCurrent graph not rendering in the web UI[^1].

This appears to be caused by some interaction between `Current.Engine`, `Lwt_preemptive`, and `OpamProcess`. The relevant section of code is https://github.com/ocurrent/opam-repo-ci/blob/master/lib/lint.ml#L99-L105, with a hopefully repeatable example below. (Note, the opam libraries are from opam being vendored in the opam-repo-ci repository.)

As `opam-repo-ci` is currently run in production on a linux machine, this is not particularly critical. However, this information may be useful if we every consider moving it to another machine.

```ocaml
Lwt_main.run @@ begin
  let f () = 
    let url = OpamFile.URL.empty in
    Lwt_io.with_temp_dir @@ fun dir ->
    Lwt_preemptive.detach begin fun () ->
      OpamProcess.Job.run @@
      OpamRepository.pull_tree "empty-pkg"
        (OpamFilename.Dir.of_string dir)
        (OpamFile.URL.checksum url)
        [OpamFile.URL.url url]
    end ()
  in
  let engine =
    Current.Engine.create (fun x -> f x |> ignore; Current.return ())
  in
  Current.Engine.thread engine
end
```

[^1]: The graph is successfully rendered by `dot` outputting the SVG to `stdout`, but the service fails to wait on `dot`'s return code, so when it exits it becomes a zombified process. `opam-repo-ci`'s rendering thread blocks until it gets the return code, so it waits forever. Relevant code is in https://github.com/ocurrent/ocurrent/blob/master/lib_web/pipeline.ml.

benmandrew:
Fixed in https://github.com/ocurrent/opam-repo-ci/pull/258.

PR #258, opened 2024-01-16
Local development pipeline

benmandrew:
`opam-repo-ci` currently has a development pipeline ([service/local.ml](https://github.com/ocurrent/opam-repo-ci/blob/master/service/local.ml)) that is nominally local, but in fact requires:
1. A GitHub repository along with GitHub authentication
2. Access to a cluster to run jobs

This PR fixes problem (1) by accessing a local Git repository through its file path, and problem (2) by using Docker to run jobs locally. This is a significant step towards supporting integration testing of the pipeline.

Various quality-of-life changes are also made to the original pipeline, for example to share code between the two. The `cluster_build` cache was originally a `Current_cache.Generic`, but as the value in the key-value pair was empty it can simply be specialised to a `Current_cache.Make`, which better represents its purpose (as far as I understand).

Depends on PR https://github.com/ocurrent/opam-repo-ci/pull/257.

PR #261, opened 2024-01-23
Support running on MacOS

benmandrew:
Fixes https://github.com/ocurrent/opam-repo-ci/issues/260.

A particular function in the linting step causes problems on MacOS, detailed in the issue above.

This PR creates a separate function that moves the opam process from a preemptive thread to the main thread. We prefer to use the original non-blocking function on Linux as it prevents an unresponsive service.

The user must specify that they are running MacOS with an optional command-line argument, `--macos`.

(Ideally we should fix the bug at the source but it looks like it'd be complete hell, so this is good enough™.)

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -118,6 +118,31 @@ module Check = struct
         | exn -> Lwt.fail exn
         end
 
+  (** On MacOS [get_dune_project_version] does not work and causes a hang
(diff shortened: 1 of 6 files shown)


benmandrew:
Included in https://github.com/ocurrent/opam-repo-ci/pull/258.

PR #262, opened 2024-01-29
Detect host OS and architecture

benmandrew:
Knowing the host OS is required to specialise the lint job, as detailed here https://github.com/ocurrent/opam-repo-ci/issues/260. This compares with the previous solution in https://github.com/ocurrent/opam-repo-ci/pull/258 that required an explicit command-line argument.

Knowing the host architecture allows the local pipeline to generate local Docker jobs that run on the correct architecture.


lib/lint.ml
```diff
@@ -295,7 +295,7 @@ module Lint = struct
 
   module Value = struct
     type t = {
-      is_macos : bool
+      host_is_macos : bool
```
tmattio, line 298: You could use a variant with the host here, instead of a boolean. That would ensure that code that needs to access the host information in the future can re-use this record.

service/pipeline.ml
```diff
@@ -7,6 +7,28 @@ module Docker = Current_docker.Default
 module Common = Opam_repo_ci_api.Common
 module Distro = Dockerfile_opam.Distro
 
+(* https://github.com/ocurrent/opam-repo-ci/issues/260 *)
+let host_is_macos =
```
tmattio, line 11: How are opam and Dune doing this? I suspect some corner cases can be covered and they would probably have been added in these projects.

lib/lint.ml
benmandrew: I've changed this to store the OS as a string instead, but storing the full variant would be a lot of work as it requires knowing how to find the distribution for every single OS as well as other information not yet necessary. (Dune has a lot of code to deal with this!)

tmattio: I'm not sure I see why a variant would be more complicated than a string. For instance, the `host_os` function can be written as:

```ocaml
let host_os =
  let os =
    Filename.quote_command "uname" ["-s"]
    |> Unix.open_process_in
    |> In_channel.input_line
    |> Option.value ~default:"Linux"
  in
  match String.lowercase_ascii os with
  | "darwin" | "osx" -> MacOS
  | s -> Other s
```

benmandrew: Ah sorry I thought you meant the platform variants in `variant.ml` (which include lots more platform information), and not OCaml variants!

Issue #264, opened 2024-01-31
Additional tests for opam-repo-ci


raphael-proust:
In https://github.com/ocaml/opam-repository/pull/25143 we realised there is a near-collision for package names: two packages where one has dashes (`-`) and the other underscores (`_`). We (the opam-repository maintainers) would like to avoid creating such near collisions. It would be good if the opam-repo-ci could add the following tests/outputs:

- When a new package file is added by a PR, and there is another package that has the same name except for non-alphanumerical characters, then there should be a warning in the CI (likely in the linting job?)

We also think it'd be nice to have additional warnings for names that are similar but not identical. Just to make typo-squatting more difficult.

- When a new package file is added by a PR, and there is another package that has a name with a low Levenshtein-distance from the newly published one, then there should be a warning in the CI. (Low would probably be 2 or 3.)

benmandrew:
For the low Levenshtein distance warning, how do you propose the value of "low" be dependent on the overall length of the package name? Without this, packages that have names 3 or 4 characters long (of which there are many) will all fail their lint checks.

raphael-proust:
The distance probably needs to only be checked after some threshold of length. I think ideally we would check the current distance distribution and then make some good guess about the appropriate threshold. As a gut feeling, I'd say the distance would be something like 1/6th. As in:

- For a package up to name-length 6, trigger a warning if there is a distance of 1.
- For a package up to name-length 12, trigger a warning if there is a distance of 2 or less.
- etc.

This might be a bit naive though… and maybe we need something fancy with a non-linear length-distance relation.

benmandrew:
I'll have a go implementing it in a general way, so that we can easily tune the threshold if there's annoying false positives.

benmandrew:
The Levenshtein distance check is not yet implemented.

PR #266, opened 2024-02-06
Check name collisions with other packages

benmandrew:
This PR adds to the linting step a check for similar package names already in the opam repository. Similarity is defined as case-insensitive equality considering `-` and `_` to be the same. As examples, by this relation:

- `lru-cache` and `lru_cache` collide
- `lru-cache` and `LRU-cache` collide
- `lru-cache` and `cache-lru` do not collide

If a similarly-named package is found a warning is surfaced. This is a partial solution to #264.

Some basic refactoring of the current lint checks is also done.


mtelvers (approved):
LGTM!

PR #270, opened 2024-02-08
Fix package name collision linting

benmandrew:
The direct usage of `opam list --all` in https://github.com/ocurrent/opam-repo-ci/pull/266 seemed to break in production, so this PR changes it so that we just walk the directory containing all of the packages ourselves, which is a lot simpler.


benmandrew:
This is being tested in production, if no issues occur I will merge.

PR #274, opened 2024-02-22
Testing for the lint check

benmandrew:
The lint checker needs a Git repository as input, so we use patches to add commits to a temporary, minimal `opam-repository` in a reproducible way.

Lint checking outside of `opam` files (e.g. for `dune` and `dune-project`) currently requires downloading the package's contents using a URL, so this requires more investigation as to how it can be tested locally.


benmandrew:
Closed in favour of #276.

PR #276, opened 2024-03-12
Cram testing for lint check

benmandrew:
Proper integration testing of the `opam-repo-ci-local` binary, replaces #274.

Currently only the lint check is tested, but this will be expanded in later PRs.

### DIFF
--- a/lib/integration_test.ml
+++ b/lib/integration_test.ml
@@ -1,0 +1,16 @@
+open Current.Syntax
+
+(* Which part of the pipeline should be tested *)
+type t =
+  | Lint
+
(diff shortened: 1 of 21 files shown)


benmandrew:
From `test/umask-022.t`:

>The lint check requires created opam files to have permissions 644.
>
>Requiring the user to change their [umask] is not ideal, it would be better to
find a way to change the permissions of the files in the cram tests. [chmod]
works on the files in the test directory, but the [opam-repo-ci-local] binary
appears to use files copied to a new directory with [umask] permissions.

punchagan (approved):

PR #280, opened 2024-03-20
Expect-test specs

benmandrew:
The list of generated specs is dumped to a file and diffed against the expected list.


mtelvers (approved):
LGTM.  Builds and runs fine.  Minor ocamlformat alert from the CI.

PR #281, opened 2024-03-21
Cram testing for package analysis step

benmandrew:
Similar to the existing testing for the lint check, we test:

- Adding a new package
- Making insignificant changes to a package
- Making significant changes to a package

We also clean up the lint testing.


mtelvers (approved):
It builds and runs for me, however there are some odd CI failures.

PR #282, opened 2024-03-25
Add detail to linting error result

benmandrew:
Allows the integration testing to check for the actual errors, rather than just the number of them.


punchagan (approved):
Ah, this makes the errors much more useful and the tests easier to debug! Thanks!

mtelvers (commented):
LGTM!

PR #283, opened 2024-03-25
List revdeps integration testing

benmandrew:
This tests that the correct reverse depencencies are listed. With our dummy `opam-repository` we add a new version of package `a-1`, on which `b.0.0.1` and `b.0.0.2` depend.

The test currently requires Docker and takes more than 2 minutes to run, as it has to go through the entire Dockerfile. The PR is a draft until this can be improved.

While running, the test pipeline can be monitored on `localhost:8080`.

Other refactoring is also included.

### DIFF
--- a/lib/cluster_build.ml
+++ b/lib/cluster_build.ml
@@ -107,9 +107,9 @@ module Op = struct
       let base = Spec.base_to_string base in
       match ty with
       | `Opam (`List_revdeps { opam_version }, pkg) ->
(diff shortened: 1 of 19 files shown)


shonfeder:
Given the number of conflicts here and the WIP to replace the CI logic for revdeps listing with a CLI tool that can list revdeps tests (and run the tests in a single worker), I'm not clear on whether we should move forward with this work. WDYT, @benmandrew and @punchagan ?

PR #286, opened 2024-03-27
Add detail to jobs-per-pr metric

benmandrew:
Splits the number of jobs per status, to give more info about what failed, succeeded, and is actually putting load on the cluster.

Screenshot 2024-03-27 at 11 14 10


punchagan (approved):
This is great! Thank you for doing this.

PR #284, opened 2024-03-26
Test package name collisions with Levenshtein distance

benmandrew:
Augments the lint check by testing whether the new package's name is within a certain edit distance of any other, principally to avoid typosquatting.

The cutoff is chosen as follows, but can be easily changed:

- For names of three characters or less, we don't check the distance
- Up to six characters, we check for an edit distance of one
- Up to twelve characters, edit distance of two
- Etc.

Closes https://github.com/ocurrent/opam-repo-ci/issues/264.

PR #292, opened 2024-04-17
Don't check name collisions for existing packages

benmandrew:
...and increase divisor from 6 to 16 to reduce false positives.

Extension of https://github.com/ocurrent/opam-repo-ci/pull/284.

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -245,7 +245,7 @@ module Check = struct
       let l = String.length p0 in
       if l <= 3 then false
       else
-        let k = ((l - 1) / 6) + 1 in
+        let k = ((l - 1) / 16) + 1 in
         Option.is_some @@ Mula.Strings.Lev.get_distance ~k p0 p1
     in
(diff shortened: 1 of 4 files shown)


benmandrew:
Fixes issue mentioned in https://github.com/ocaml/opam-repository/pull/25678#issuecomment-2060638622.

hannesm:
Thanks for your work. As far as I observe, this does 95% of what the title advertises:
if there's a single version released to opam repository, and this one is modified, the check kicks in.

So, if we have foo.0.1.0 released, and someone modifies it (due to upper bounds or whatever), this check for name collisions is run when it should not.

I'm not sure how an appropriate fix would be, since you'll need the old git state vs the new git state, and peek whether in the old state that package was already present.

shonfeder:
The rest of that logic should be taken care of in https://github.com/ocurrent/opam-repo-ci/pull/316, if I'm not mistaken.

We are just waiting on final approval for the prefix logic before merging and deploying that.

Issue #293, opened 2024-04-18
Service unreliability


benmandrew:
The `opam-repo-ci` service has had random and inconsistent downtime for a long time now, as exhibited in the three-month uptime graph below.

<img width="508" alt="Screenshot 2024-04-18 at 13 19 57" src="https://github.com/ocurrent/opam-repo-ci/assets/13054139/7ab641af-6213-45a4-9d24-5983a072093f">

When these outages occur, the service and web-ui stop sending metrics and responding to requests. However, the machine still sends its own metrics, showing a single CPU core pinned to 100% utilisation (of which ~75% is user-thread and ~25% is system-thread), implying that the service is spinning on something.

When I run `opam-repo-ci-service` locally, I see lots of warnings logged. There seems to be one for each OCluster job that runs to completion. This reference leaking may be connected to the unreliability issues. After a certain point, I am also unable to Ctrl-C the program and have to force-kill it.
```
capnp-rpc [WARNING] Reference GC'd with rc=1!
                    switchable(4394) (unset, rc=1) -> far-ref(4393) -> i544
```

These issues are not seen in `opam-repo-ci-local`. The main difference between the two is that `-local` uses local Docker containers to run jobs, while `-service` uses a connection to an external cluster scheduler using OCluster. Thus, it is likely that the issue arises here.

talex5:
> These issues are not seen in opam-repo-ci-local. The main difference between the two is that -local uses local Docker containers to run jobs, while -service uses a connection to an external cluster scheduler using OCluster. Thus, it is likely that the issue arises here.

`opam-repo-ci-local` used to work with a cluster. It looks like you removed that feature in #258.

The PR comment said it required access to a cluster to run jobs. This is true; however, the "cluster" can be the local machine (the `docker-compose.yml` shows how to do this).

Anyway, if you have access to the machine running the service, running strace on it would be useful to see what it's doing.

shonfeder:
We talked about this in our team meeting today. I'll do my best to summarize from my notes, but it may need correction, as I am still onboarding and lacking a lot of context:

@mtelvers noted that he reset the SQLite database late on the 17th, and @benmandrew pointed out that this roughly corresponds with the recovery and stabilization of the service since the 18th. Mark proposed the following hypothesis:

- The index database is never pruned. 
- As the tables grow, perhaps an expensive query begins to eat up a lot of time.
[...]

PR #295, opened 2024-04-23
Add logs to all SQL queries

benmandrew:
Supporting the investigation into https://github.com/ocurrent/opam-repo-ci/issues/293.


mtelvers (approved):
LGTM.  My money is on `get_jobs`.  In just the few days since the database was cleared the tables in the JOIN are several hundred thousand rows.  Let's wait and see!

```
sqlite> SELECT COUNT(*) from cache;
138866
sqlite> SELECT COUNT(*) from ci_build_index;
191524
```

PR #296, opened 2024-04-23
Update submodules to include Fedora 40

mtelvers:


shonfeder (approved):
LGTM!

Thanks for point this out :)

Am I correct in understanding that the `ocaml-dockerfile` update is what's motivating the update, and the updates to `ocluster` and `ocurrent` are just picking up changes from the slightly out of date (git-submoduled) deps?

mtelvers:
Yes, exactly that.

shonfeder:
Thank you for confirming! Could you update the PR title so it mentions the motivating Fedora update (just to make the trail a bit clearer in case we need to look back).

PR #303, opened 2024-05-09
Remove redundant --installable flag alongside --coinstallable-with

punchagan:
The --installable flag runs a filter using the installable_subset function, which is just a call to the coinstallable_subset function with an empty set as the coinstallable set of packages. The other argument of --coinstallable-with already runs the same filter with the specified package as the set with which need to be coinstallable.


punchagan:
> Looks good to me, could you update the expect tests?

Updated! Apologies for missing them on the original commit. 

(Also, pushed an unrelated minor typo fix commit, separately)

benmandrew (approved):

PR #304, opened 2024-05-10
Lint: ensure checksum is present and not MD5

hannesm:
Dear valued developers,

I hereby propose another lint check for the amazing opam-repo-ci -- namely something that ensures that all artifacts include a checksum that is not MD5. MD5 is since years not a secure digest, but forgeable. We should ensure the opam package system is not extended with more packages that include this weak hash algorithm.

Since I have not succeeded to compile opam-repo-ci locally, I'd await your (or the CI) feedback on this change.

Also, since this is a new check, I'm happy to hear your opinion about it.

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -26,6 +26,7 @@ type error =
   | OpamLint of (int * [`Warning | `Error] * string)
   | FailedToDownload of string
   | NameCollision of string
+  | WeakChecksum of string
 
 type host_os = Macos | Other [@@deriving to_yojson]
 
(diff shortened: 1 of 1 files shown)


benmandrew:
Looks good to me, thanks!

Just to clarify, we'd still like to allow MD5, just not without SHA-256 or -512?

hannesm:
> Looks good to me, thanks!
> 
> Just to clarify, we'd still like to allow MD5, just not without SHA-256 or -512?

That's what the current PR does, exactly. This is based on the observation that opam checks all provided checksums (and not only one). I'd be happy if someone could independently verify that this is the case (I did so more than a year ago, leading to https://opam.ocaml.org/blog/opam-2-1-5-local-cache/).

I'd as well be fine to have a stricter check, but I'd guess we should adapt opam-publish first to not emit md5 (and sha512) checksums, but only emit sha512.

benmandrew:
Looks like it does check every single one, so no harm in including MD5 if stronger hashes are present. Thank you for the contribution!

PR #306, opened 2024-05-14
New lint: disallow pin-depends to be present.

hannesm:
The opam-repository should not carry any pin-depends. Sometimes such opam packages are submitted since they come from development repositories where related libraries need to be pinned. It is tedious (and error-prone) to rely on humans checking for the not-occurence of pin-depends.

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -27,6 +27,7 @@ type error =
   | FailedToDownload of string
   | NameCollision of string
   | WeakChecksum of string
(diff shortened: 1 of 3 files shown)


benmandrew (approved):
Looks great, thanks!

hannesm:
Thanks. Any chance this and the earlier check could be pushed to production?

benmandrew:
Done. Let me know if you see any issues.

PR #314, opened 2024-05-23
Document tested platforms

benmandrew:
Closes https://github.com/tarides/infrastructure/issues/328

Produce user-friendly documentation for the set of tested platforms. This could be derived from `test/specs.expected` before but that file is very noisy.

Docs are produced by `dune build @doc` and if there is a difference with the existing, promoted by `dune promote`.

I don't know how common it is to look at the raw markdown table, but further work could add padding to cells so that it's more readable raw.


shonfeder (changes requested):
Sorry I didn't see this earlier!

This code reads beautifully and it's a really nice, simple solution to meet the need. I think it would be great to update the Discuss question with a link to the generated doc!

I worked out my suggestions thru code in the PR https://github.com/benmandrew/opam-repo-ci/pull/1 into your branch. If you find any of those suggestions off base maybe we can discuss? But you like the spirit of the suggestion but not the implementation, please feel free to disregard the particulars of the PR.

The one other suggestion I had was to add a link to the generated documentation from the readme?

benmandrew:
I agree with all the suggestions, thanks!

shonfeder (approved):
Thank you! Looks great :D

PR #316, opened 2024-05-27
Add linting check for restricted prefixes

### LINKED ISSUE #315: Add linting check for new restricted prefixes
Following https://github.com/ocaml/opam-repository/pull/25861, the opam repository aims to treat the following package name prefixes as restricted:

- `arch-`
- `system-`
- `host-`

The need linting rules added to the CI checks on opam files to raise an alert when package names start with one of these prefixes.

cc @raphael-proust and @kit-ty-kate

benmandrew:
Fixes #315.


raphael-proust:
The list of conflict-class (that we want the linter to help keep in sync with the prefixes) is discussed on https://github.com/ocaml/opam-repository/pull/25861#pullrequestreview-2057999557

The discussion is not complete (as time of writing of this comment).

benmandrew:
The current prefix to conflict class mapping according to my understanding is:

- `arch-` to `ocaml-architecture-setting`
- `system-` to `ocaml-system-setting`
- `host-arch-` to `ocaml-host-archicture`
- `host-system-` to `ocaml-host-system`
- `msys2-` to `msys2-environment`

This can easily be changed as the discussion evolves.

lib/lint.ml
```diff
@@ -383,6 +453,9 @@ module Check = struct
   let of_dir ~host_os ~master ~job ~packages cwd =
     let master = Current_git.Commit.hash master in
     exec ~cwd ~job [|"git"; "merge"; "-q"; "--"; master|] >>/= fun () ->
+    get_packages ~cwd >>= fun existing_packages ->
+    prev_versions_exist ~cwd packages existing_packages >>= fun prev_versions_exist ->
+    let is_new_package = not prev_versions_exist in
```
shonfeder, line 458: Iiuc, this will be true if any single package in a PR is new, but couldn't that end up meaning we get several unhelpful checks linting checks if, for instance, one new brand new package is released among several pre-existing packages? E.g., say a new package is factored out of `lwt`, called `lwt-new`, and released along with updates to the `lwt` package. Since the latter didn't exist before, and we are here checking all `packages`, `is_new_package` will be `true`. But then, iiuc, when we linting the `lwt` package, we will end up checking it for name collisions. If my reading of the logic is correct, then it is probably better to move this test into the fold.

Iiuc, a function along these lines should be able to give us a constant time check for whether a single package is being published for the first time:

```ocaml
  let is_newly_published_package ~cwd ~job package master =
    let package_name = OpamPackage.Name.to_string package.OpamPackage.name in
    (* [git cat-file -e branch:path] will exit with zero status if the object at
       [path] exists on [branch] *)
    exec ~cwd ~job [|"git"; "cat-file "; "-e"; master^":"^package_name|]
    >|= function
    | Error _ -> true (* The package directory does not exist on master *)
    | Ok _ -> false (* The package directory does exist on master *)
```

And it could be used as the check for a newly added package on line 475 with something like

```ocaml
          is_newly_published_package ~cwd ~job pkg master >|=
          function
          | false -> errors
          | true ->
              let errors = check_name_collisions ~errors ~pkg existing_packages in
              Prefix.check_name_restricted_prefix ~errors ~pkg
```

In case you like this approach, you can use or reference the patch here: https://github.com/ocurrent/opam-repo-ci/commit/5b67c0a12540ac85932b50906b4e45c0401ce489

test/lint.t
```diff
@@ -28,10 +29,12 @@ Tests the following:
...
   * a-1 (master)
   $ opam-repo-ci-local --repo="." --branch=new-branch-1 --lint-only --no-web-server
-  Error "3 errors:
+  Error "5 errors:
   Error in b.0.0.1:            warning 25: Missing field 'authors'
   Error in b.0.0.2:              error  3: File format error in 'unknown-field' at line 11, column 0: Invalid field unknown-field
-  Error in b.0.0.3: pin-depends present. This is not allowed in the opam-repository."
+  Error in b.0.0.3: pin-depends present. This is not allowed in the opam-repository.
+  Error in system-b.0.0.1: package with prefix 'system-' requires conflict class 'ocaml-system-setting'
+  Warning in system-b.0.0.1: package name has restricted prefix 'system-'"
```
shonfeder, line 37: Can you also include a test for the other conflict class error? `"Error in %s: package with conflict class '%s' requires name prefix '%s'"`

lib/lint.ml
```diff
@@ -437,67 +517,79 @@ module Lint = struct
...
-        Fmt.str "Error in %s: Weak checksum algorithm(s) provided. Please use SHA-256 or SHA-512. Details: %s" pkg msg
-      | PinDepends ->
-        Fmt.str "Error in %s: pin-depends present. This is not allowed in the opam-repository." pkg
-    )
+  let msg_of_prefix_conflict_class_mismatch ~pkg = function
+    | WrongPrefix (conflict_class, required_prefix) ->
+      Fmt.str
+        "Error in %s: package with conflict class '%s' requires name prefix '%s'"
+        pkg conflict_class required_prefix
+    | WrongConflictClass (prefix, required_conflict_class) ->
```
shonfeder, line 525: The names used in these patters would be nice choices for the record field names on the type, if you chose to go with those.

shonfeder (commented):
Looking good! I have one question and related suggestion about the new-package detection logic. My other suggestions are minor documentation and testing-related things.

benmandrew:
Recommendations have been integrated and the PR has been rebased onto master; now we're waiting for the finalisation of the discussion on `ocaml/opam-repository`.

shonfeder:
> The list of conflict-class (that we want the linter to help keep in sync with the prefixes) is discussed on https://github.com/ocaml/opam-repository/pull/25861#pullrequestreview-2057999557
> 
> 
> 
> The discussion is not complete (as time of writing of this comment).

@raphael-proust could you confirm whether the prefix/conflict classes here match what you are expect from a maintenance angle?

raphael-proust:
The prefix (presented as a glob) / conflict classes (presented as a literal string) that were settled on are:

```
mysys2-* / "msys2-env"
arch-* / "ocaml-arch"
ocaml-env-mingw* / "ocaml-env-mingw"    -- NOTE: the globbing pattern for file names doesn't have a dash
ocaml-env-msvc* / "ocaml-env-msvc"      -- NOTE: same
host-arch-* / "ocaml-host-arch"
host-system-* / "ocaml-host-system"
system-* / "ocaml-system"
```

shonfeder:
@punchagan since I've touched this a fair bit and you're looking at the linting stuff anyhow, would mind reviewing this when you have a chance?

shonfeder (approved):

PR #318, opened 2024-05-29
Document the `--lint-only` flag for local use

benmandrew:
Also make the README a bit more readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# opam-ci
+# opam-repo-ci
 
 Status: **experimental**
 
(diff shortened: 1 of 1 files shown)


benmandrew:
Unfortunately the lint check is tightly coupled to OCurrent, so we have to spin up a full pipeline and bail when we hit the right point. This means that if an error occurs prior to the lint check (e.g. in analysis), the program will not exit and user will have to inspect it themselves. Ideally the check should be able to run independently of OCurrent.

shonfeder (approved):
Sorry for the delay here! I forgot to push "submit". Looks great to me. Thank you for the updates to the docs!

PR #320, opened 2024-05-30
Move all vendored git submodule to pin-depend

shonfeder:
Closes tarides/infrastructure#320

Replaces all vendored git-submodules with pin-depends.

This enable us to:

- Install and build switches just using opam, without needing special installation instructions.
- Will make it easy to switch to published versions of its deps (which we'd like to do).
- Will make it easy to move the service into a service-monorepo, if we end up deciding to do that.

Due to a quirk in ocaml-ci, we cannot just pin to master, so updates here will require a find/replace of the hash of the package to be updated. This is only slightly more cumbersome than a git sub-module update. I'd like to fix this in a followup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,24 +2,8 @@ FROM ocaml/opam:debian-12-ocaml-4.14@sha256:06d58a5cb1ab7875e8d94848102be43f6492
 RUN sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam && opam init --reinit -ni
 RUN sudo apt-get update && sudo apt-get install libev-dev capnproto graphviz m4 pkg-config libsqlite3-dev libgmp-dev libffi-dev -y --no-install-recommends
 RUN cd ~/opam-repository && git fetch origin master && git reset --hard 6d3d7021d944058cc7531905e065fe2cd72f01e8 && opam update
-COPY --chown=opam \
-	ocurrent/current_docker.opam \
-	ocurrent/current_github.opam \
-	ocurrent/current_git.opam \
-	ocurrent/current.opam \
-	ocurrent/current_rpc.opam \
-	ocurrent/current_slack.opam \
-	ocurrent/current_web.opam \
-	/src/ocurrent/
-WORKDIR /src
-RUN opam pin add -yn current_docker.dev "./ocurrent" && \
-    opam pin add -yn current_github.dev "./ocurrent" && \
-    opam pin add -yn current_git.dev "./ocurrent" && \
-    opam pin add -yn current.dev "./ocurrent" && \
-    opam pin add -yn current_rpc.dev "./ocurrent" && \
-    opam pin add -yn current_slack.dev "./ocurrent" && \
-    opam pin add -yn current_web.dev "./ocurrent"
 COPY --chown=opam opam-repo-ci-service.opam opam-repo-ci-api.opam /src/
+WORKDIR /src
 RUN opam install -y --deps-only .
 ADD --chown=opam . .
 RUN opam exec -- dune build ./_build/install/default/bin/opam-repo-ci-service
@@ -30,7 +14,7 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add -
 RUN echo 'deb https://download.docker.com/linux/debian bookworm stable' >> /etc/apt/sources.list
 RUN apt-get update && apt-get install docker-ce -y --no-install-recommends
 RUN git config --global user.name "ocaml" && git config --global user.email "ci"
-WORKDIR /var/lib/ocurrent
+WORKDIR /
 ENTRYPOINT ["dumb-init", "/usr/local/bin/opam-repo-ci-service"]
 ENV OCAMLRUNPARAM=a=2
 COPY --from=build /src/_build/install/default/bin/opam-repo-ci-service /usr/local/bin/
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add -
 RUN echo 'deb https://download.docker.com/linux/debian bookworm stable' >> /etc/apt/sources.list
 RUN apt-get update && apt-get install docker-ce -y --no-install-recommends
 RUN git config --global user.name "ocaml" && git config --global user.email "ci"
-WORKDIR /
+WORKDIR /var/lib/ocurrent
 ENTRYPOINT ["dumb-init", "/usr/local/bin/opam-repo-ci-service"]
 ENV OCAMLRUNPARAM=a=2
 COPY --from=build /src/_build/install/default/bin/opam-repo-ci-service /usr/local/bin/
--- a/dune
+++ b/dune
@@ -1,3 +1,3 @@
 (dirs :standard \ var)
 
-(vendored_dirs ocluster ocaml-dockerfile ocaml-version opam)
+(vendored_dirs ocaml-dockerfile ocaml-version opam)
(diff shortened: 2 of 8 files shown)


opam-repo-ci-service.opam
```diff
@@ -39,12 +38,51 @@ depends: [
...
+
+  ["ocluster-api.0.3.0~dev"      "git+https://github.com/ocurrent/ocluster.git#912ecfd38f003c6f537abca7e5313973b15ba615"]
+  ["current_ocluster.0.3.0~dev"  "git+https://github.com/ocurrent/ocluster.git#912ecfd38f003c6f537abca7e5313973b15ba615"]
+
+  ["dockerfile.8.2.1~dev"       "git+https://github.com/ocurrent/ocaml-dockerfile.git#85891cad42189a03960b48792293e8fb3a18552c"]
+  ["dockerfile-opam.8.2.1~dev"  "git+https://github.com/ocurrent/ocaml-dockerfile.git#85891cad42189a03960b48792293e8fb3a18552c"]
+
+  ["ocaml-version.3.6.7~dev"  "git+https://github.com/ocurrent/ocaml-version.git#b80aad5425c7dd6f1e6f8925f5719c72f8b8f7fa"]
+
+  ["obuilder-spec.0.6.0~dev" "git+https://github.com/ocurrent/obuilder.git#c0eb046f52ca0c5fd9b560d4bdc0cb11c852d4f7"]
```
shonfeder, line 83: We wouldn't need commit hashes on all these, but ocam-ci has a constraint requiring a has. We can fix this in a followup.

See https://github.com/ocurrent/ocaml-ci/blob/a29bd00fcea21bcbcc1876d693b356e84f9c1236/lib/pin_depends.ml#L96-L102

shonfeder:
The CI failures are just [debian-12-4.14_x86_32_opam-2.1 (failed: Failed: Build failed)](https://ocaml.ci.dev/github/ocurrent/opam-repo-ci/commit/d8ac120af81c2d22f03000ec3dc170c1d12faf21/variant/debian-12-4.14_x86_32_opam-2.1/-/refs/heads/shon/infra-320/vendors-to-pins), which are not be filtered out as I'd expect given https://github.com/ocurrent/ocaml-ci/pull/932, even tho it *is* being filtered out for ocaml-ci itself! :disappointed: -- but that is besides the point.

Dockerfile
```diff
@@ -30,7 +14,7 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add -
 RUN echo 'deb https://download.docker.com/linux/debian bookworm stable' >> /etc/apt/sources.list
 RUN apt-get update && apt-get install docker-ce -y --no-install-recommends
 RUN git config --global user.name "ocaml" && git config --global user.email "ci"
-WORKDIR /var/lib/ocurrent
+WORKDIR /
```
mtelvers, line 17:
```suggestion
WORKDIR /var/lib/ocurrent
```
We should not change this line.

This `WORKDIR` becomes the current working directory in the Docker container at runtime.  This is the mount point we use for the data volume, storing the database and job logs, etc.

mtelvers:
> The CI failures are just [debian-12-4.14_x86_32_opam-2.1 (failed: Failed: Build failed)](https://ocaml.ci.dev/github/ocurrent/opam-repo-ci/commit/d8ac120af81c2d22f03000ec3dc170c1d12faf21/variant/debian-12-4.14_x86_32_opam-2.1/-/refs/heads/shon/infra-320/vendors-to-pins), which are not be filtered out as I'd expect given [ocurrent/ocaml-ci#932](https://github.com/ocurrent/ocaml-ci/pull/932), even tho it _is_ being filtered out for ocaml-ci itself! 😞 -- but that is besides the point.

There are some CI failures on debian-12 which aren't related to the x86_32 issue.  e.g. https://ocaml.ci.dev/github/ocurrent/opam-repo-ci/commit/d8ac120af81c2d22f03000ec3dc170c1d12faf21/variant/debian-12-5.2_arm64_opam-2.1.

I think some will go away with a `dune build --auto-promote`

mtelvers:
I also note that this change significantly increases the build time on my system.
```
$ git checkout master && rm -rf _build/ && time dune build
Switched to branch 'master'
Your branch is up to date with 'upstream/master'.
schema.capnp --> schema.mli schema.ml
ocurrent.capnp --> ocurrent.mli ocurrent.ml
kj/filesystem-disk-unix.c++:1690: warning: PWD environment variable doesn't match current directory; pwd = /home/mtelvers/opam-repo-ci
                                        
real	0m5.771s
user	0m22.141s
sys	0m5.079s
```
Using this PR.
```
$ git checkout pr320 && rm -rf _build/ && time dune build
Switched to branch 'pr320'
schema.capnp --> schema.mli schema.ml
ocurrent.capnp --> ocurrent.mli ocurrent.ml
kj/filesystem-disk-unix.c++:1690: warning: PWD environment variable doesn't match current directory; pwd = /home/mtelvers/opam-repo-ci
                                        
real	0m26.479s
user	0m41.993s
sys	0m6.208s
```

benmandrew (approved):
Approved with one change.

Dockerfile
shonfeder: Got it. I can't recall why I changed this before, and I've reverted it.

dune
```diff
@@ -1,3 +0,0 @@
-(dirs :standard \ var)
```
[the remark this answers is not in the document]
mtelvers: When you run the service it writes to `var/git`, `var/db` of the current working directory.  Hence my comments about `WORKDIR` in the Dockerfile.  @benmandrew and I have probably run the service locally on our machines for development.  For me, `~/opam-repo-ci$ find var/ | wc -l` shows I have >200K files.  Reinstating the file does fix the build time issue I was seeing.

shonfeder:
@mtelvers I am not seeing long build times. But maybe it if from the `var` file @benmandrew mentioned. I've excluded that directory from the dune dirs again, maybe that will sort things out?

Worth noting I guess that the CI systems don't seem to showing long build times afaict? 

> There are some CI failures on debian-12 which aren't related to the x86_32 issue. e.g. https://ocaml.ci.dev/github/ocurrent/opam-repo-ci/commit/d8ac120af81c2d22f03000ec3dc170c1d12faf21/variant/debian-12-5.2_arm64_opam-2.1.
> 
> I think some will go away with a `dune build --auto-promote`

Is it possible those are results from an earlier commit? Later commits aren't showing these testing failures, and I don't know how that test could be failing on just one platform?

I do see some debian-12s were failing with package installation trouble, but that is not recurring on rebuild, so seems like it may be flakiness?

mtelvers:
Yes, I did comment on @benmandrew thread that the `var` exclusion for Dune was indeed the issue.

CI looks clean with the current build.

shonfeder:
Are you all OK with me merging this, or any concerns that it may not deploy correctly etc.? Anything else you'd suggest to doing to reduce the chance of any outage from deploying this change?

shonfeder:
Ok! Will do. Thanks :)

shonfeder:
This is stalled out due to:

- CI failures due to surprising dependency resolution errors that I need to debug
- The limitations imposed on pin-depends by ocaml-ci make the benefits of this change marginal, at best (and maybe even makes it a net cost), but I don't think it makes sense for us to invest in the needed changes to ocaml-ci given our current objectives.
- Given recent discussions around ways we want to refactor the services, this seems lower priority.
- What we really want is to be using published versions of these dependencies. A positive pressure there is that it will force us to share changes to the stack, and give us good feedback on the pains of getting packages published when needed.

shonfeder:
I've become more convinced that a mono-repoization is gonna be more useful here. In any case, until the limitation with ocaml-ci is fixed, this will create more (rather than less) work to maintain.

PR #330, opened 2024-07-01
Fix tests

mtelvers:
I inadvertently pushed my last commit to master.

This commit updates the test results to work with the latest ocaml-dockerfile submodule.

### DIFF
--- a/doc/platforms.md
+++ b/doc/platforms.md
@@ -2,24 +2,20 @@
 
 ## Operating Systems
 
-- alpine-3.19
+- alpine-3.20
 - archlinux
-- debian-10
 - debian-11
 - debian-12
 - debian-testing
 - debian-unstable
-- fedora-38
 - fedora-39
 - fedora-40
 - freebsd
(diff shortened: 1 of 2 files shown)


shonfeder:
Looks like another promotion is needed to satisfy the linting check: https://ocaml.ci.dev/github/ocurrent/opam-repo-ci/commit/edca6c4ba1a15f49da22c5d7eff19b0fa785a454/variant/%28lint-doc%29#L1028-1028

punchagan:
Looks like the change to deprecate Ubuntu 23.10 in https://github.com/ocurrent/ocaml-dockerfile/pull/214 which is due to be merged next week is already a part of the submodule on master. Do we want to revert it? Or are we okay to deprecate the OS a week earlier?

mtelvers:
I'm happy to deprecate it a week early so save an extra redeployment.

punchagan (approved):
Looks good. Thanks!

PR #313, opened 2024-05-23
Add a new lint check: disallow the extra-files field in opam packages

hannesm:
There's a great alternative - extra-sources - that should be used instead.

The motivation is https://github.com/ocaml/opam-repository/issues/25876, and this PR should be merged and put live once opam-repository is in the new state. I'll comment once this is the case.

I open this PR for (a) CI run (b) review, and (c) tracking the progress of https://github.com/ocaml/opam-repository/issues/25876

PR #336, opened 2024-07-04
Remove hash check for extra files

punchagan:
d6df6dbfd553750b9c1d90ab06442a6ae93fee8b in #313 added a check to prevent use of extra-files in the opam-repository. Also, the repository doesn't contain any existing packages with extra-files. This means, we can get rid of the hash check for extra files in the linter.

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -568,7 +568,7 @@ module Lint = struct
                   but your opam file only requires dune >= %s. Please check which requirement is the right one, and fix the other."
           pkg ver dep
     | UnexpectedFile file ->
-        Fmt.str "Error in %s: Unexpected file in %s/files/%s" pkg (Check.path_from_pkg package) file
+        Fmt.str "Error in %s: Unexpected file in %s/%s" pkg (Check.path_from_pkg package) file
     | ForbiddenPerm file ->
         Fmt.str
           "Error in %s: Forbidden permission for file %s/%s. All files should have permissions 644."
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -91,32 +91,18 @@ module Check = struct
   let scan_dir ~cwd errors pkg =
     let dir = Fpath.to_string cwd // path_from_pkg pkg in
     get_files dir >>= fun files ->
-    let rec aux errors extra_files = function
-      | [] -> Lwt.return (errors, extra_files)
+    let rec aux errors  = function
+      | [] -> Lwt.return errors
       | "opam"::files ->
           is_perm_correct (dir // "opam") >|= begin function
           | true -> errors
           | false -> ((pkg, ForbiddenPerm (dir // "opam")) :: errors)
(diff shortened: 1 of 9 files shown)


shonfeder (approved):
Looks great! Thanks for taking care of this!

punchagan:
Thanks for the review. Pushed a change fixing some extra spaces and parens - I'm spoiled by ocamlformat.

punchagan:
> Looks great! Thanks for taking care of this!

@shonfeder I don't see any packages in the repository with a `files/` directory. I'm not sure `files/` directory can contain any other files, if we are discouraging the use of `extra-files` for patches to a package.  If `files/` is only meant for patches, we can also improve the lint [here](https://github.com/punchagan/opam-repo-ci/blob/remove-extra-files-hash-check/lib/lint.ml#L102).

punchagan:
> @shonfeder I don't see any packages in the repository with a `files/` directory. I'm not sure `files/` directory can contain any other files, if we are discouraging the use of `extra-files` for patches to a package. If `files/` is only meant for patches, we can also improve the lint [here](https://github.com/punchagan/opam-repo-ci/blob/remove-extra-files-hash-check/lib/lint.ml#L102).

Looking at the [files/](https://opam.ocaml.org/doc/Manual.html#files) and [extra-files](https://opam.ocaml.org/doc/Manual.html#opamfield-extra-files) sections, it looks like we could change the lint to raise an error if the `files/` directory is present. But, I'm not entirely sure and it'd be nice to get a confirmation and implement the change.

shonfeder:
I'm afraid I'm not sure on the optimal behavior here myself. Let's ask the more seasoned opam repo maintainers next week.

punchagan:
@hannesm Given your change in #313 to prevent the use of `extra-files`, is the `files/` directory no longer allowed to be present in the package? Or can it be used for files which are not patches, but something else?

hannesm:
Hey, thanks for taking care of this. Indeed the `files` directory must not be present anymore, since the usage of `extra-files` is no longer allowed for opam-repository.

punchagan:
> Indeed the `files` directory must not be present anymore, since the usage of `extra-files` is no longer allowed for opam-repository.

Thanks for confirming, @hannesm !  I've updated this PR to reflect this.

shonfeder:
Adding the fix in https://github.com/ocurrent/ocurrent-deployer/commit/b60799f2acae3ee47e5c152c45cdcbf243d1d2d1 thru an analogous template file here should fix the CI :)

shonfeder (approved):

punchagan:
> Adding the fix in [ocurrent/ocurrent-deployer@b60799f](https://github.com/ocurrent/ocurrent-deployer/commit/b60799f2acae3ee47e5c152c45cdcbf243d1d2d1) thru an analogous template file here should fix the CI :)

@shonfeder I've added the fix to get CI green.

PR #337, opened 2024-07-05
Improve test to verify is_newly_published works

punchagan:
4d94c9f5437050e8007d9f6604d53c5b7593ca62 fixed an extra space in the shell command being run to detect new packages, which caused the exec to crash and caused the lint to fail. Our cram tests should have caught this, but didn't. This is because the spawned git command is expected to fail when run for a newly published package and the name conflict test only tests with newly published packages, and no already published ones.

This commit adds an "initial" package (field1) to the name collision test, which shouldn't trigger a check for name collisions. This test improvement would help catch any future regressions, even when extract the linting code to a separate tool or library.

NOTE: It is weird that the package `fielf` conflicts with `field1`, but not with `field`. This weirdness can be confusing to end-users, but is out of the scope of this commit.

### DIFF
--- a/test/lint.t
+++ b/test/lint.t
@@ -59,7 +59,7 @@ Delete OCurrent cache
   $ rm -rf var/
 
 Tests the package name collisions detection by adding initial
-packages [field] and [fieldfind] to master, and new packages
+packages [field], [field1] and [fieldfind] to master, and new packages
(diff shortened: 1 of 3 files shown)


shonfeder (approved):
Fantastic. Thank you for taking care of this followup! Looks great :)

PR #342, opened 2024-07-15
Increase timeout

mtelvers:
As a quick workaround, try a much longer timeout.
https://github.com/ocaml/infrastructure/issues/147

PR #343, opened 2024-07-17
Use 0install solver for revdeps

mtelvers:
Addressing https://github.com/ocaml/infrastructure/issues/147, I have previously proposed #342 to work around this issue by increasing the timeout for the solver, but as an alternative, this PR switches the solver for the revdeps from the default to 0install.

We do not want to change the default solver globally as we want opam-repo-ci to mirror the user's experience, but for revdeps we are more interested in whether this was a breaking change.

In my testing, the 0install solver was able to find a solution in a few seconds, whereas the default solver failed to find a solution in > 8 minutes (500 seconds).  I note that `opam` consumes 100% CPU for those 8 minutes, so this change should also have a positive effect on the time taken to run the jobs.

As an aside, I have compiled a collection of all the solver timeouts we have seen this year.  https://github.com/mtelvers/solver-timeouts

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -9,7 +9,7 @@ let cache ~variant =
                 Obuilder_spec.Cache.v "homebrew" ~target:"/Users/mac1000/Library/Caches/Homebrew" ]
 let network = ["host"]
 
-let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~pkg =
+let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~revdep ~pkg =
   let pkg_s = OpamPackage.to_string pkg in
   let with_tests_opt = if with_tests then " --with-test" else "" in
   let cache = cache ~variant in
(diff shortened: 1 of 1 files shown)


shonfeder (approved):
LGTM.

Awesome that you dug in to find a cleaner fix than just bumping the timeout again. Not only does it sound like this will address the timeout problem, but it will also yield a more efficient system overall.

mseri:
The CI seems to work fine overall, however opam 2.0 checks are failing:
```
The only failing test target in the opam-ci suite is opam-2.0-ocaml-5.2. It fails for a reason seemingly unrelated to OCANNL. I could depend on opam >= 2.1 going forward.

/home/opam: (run (shell "opam option solver=builtin-0install"))
opam: unknown command `option'.
```

Is there any particular reason we still want to keep opam 2.0 in the test matrix?

mtelvers:
@mseri Are those the errors that you are seeing related to https://github.com/ocurrent/opam-repo-ci/pull/351 which was merged today and then reverted?

shonfeder:
Yes, this is the regression noted in #352

PR #344, opened 2024-07-19
Move newly published package detection to analysis

punchagan:
Previously, we had code to detect newly published packages in the lint
phase to be able to run additional checks for such packages. But, this
detection belongs in the analysis phase, really.

Additionally, this PR also clarifies that the New kind really means NewRelease.

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -26,8 +26,9 @@ module OpamPackage = struct
 end
 
 module Analysis = struct
+  type change = Release [@@deriving eq, yojson]
   type kind =
-    | New
+    | New of change
     | Deleted
     | Unavailable
     | SignificantlyChanged
@@ -110,18 +111,18 @@ module Analysis = struct
   let add_pkg ~path ~name ~package kind pkgs =
     let update old_kind = match old_kind, kind with
       (* NOTE: Impossible combinations (opam file would have to be processed more than once) *)
-      | (New | Deleted | Unavailable), (New | Deleted | Unavailable) ->
+      | (New Release | Deleted | Unavailable), (New Release | Deleted | Unavailable) ->
           old_kind (* old_kind instead of assert false because OpamStd.Map.update works this way :( *)
       (* NOTE: stronger_kind >= weaker_kind *)
-      | New, (SignificantlyChanged | InsignificantlyChanged)
+      | New Release, (SignificantlyChanged | InsignificantlyChanged)
       | Deleted, (SignificantlyChanged | InsignificantlyChanged)
       | Unavailable, (SignificantlyChanged | InsignificantlyChanged)
       | SignificantlyChanged, (SignificantlyChanged | InsignificantlyChanged)
       | InsignificantlyChanged, InsignificantlyChanged ->
           old_kind
       (* NOTE: weaker_kind < stronger_kind *)
-      | SignificantlyChanged, (New | Deleted | Unavailable)
-      | InsignificantlyChanged, (New | Deleted | Unavailable | SignificantlyChanged) ->
+      | SignificantlyChanged, (New Release | Deleted | Unavailable)
+      | InsignificantlyChanged, (New Release | Deleted | Unavailable | SignificantlyChanged) ->
           kind
     in
     OpamPackage.Map.update (get_package_name ~path ~name ~package) update kind pkgs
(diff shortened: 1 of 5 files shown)


mtelvers (commented):
Since `NewPackage` and `NewRelease` are treated equally in virtually every case, I wonder if it would be cleaner to write

```
type change = | Package | Release
type kind = | New of change | Deleted | Unavailable | ...
```
Then in most cases, the change would be (eg):

```
-     | (New | Deleted | Unavailable), (New | Deleted | Unavailable) ->
+     | (New _ | Deleted | Unavailable), (New _ | Deleted | Unavailable)
```
And then when it is different you would have
```
match kind with
| New Package -> etc
| _ -> Lwt.return errors
```

lib/analyse.ml
```diff
@@ -191,7 +195,14 @@ module Analysis = struct
...
               | Error _ ->
                 (* new release *)
-                Lwt_result.return (add_pkg ~path ~name ~package New pkgs)
+                let package_dir = "packages/"^name in
+                let cmd = "", [| "git"; "show"; master^":"^package_dir |] in
+                Current.Process.check_output ~cwd:dir ~cancellable:true ~job cmd >>= begin function
+                | Ok _ -> Lwt_result.return (add_pkg ~path ~name ~package NewRelease pkgs)
+                | Error _ ->
+                  (* newly published package *)
+                  Lwt_result.return (add_pkg ~path ~name ~package NewPackage pkgs)
```
shonfeder, line 204: I suggest retaining the named helper function from the original code in lint, as it makes the function and intent of the logic here much easier to scan IMO. Since we are only checking for `Ok _` or `Error _` without any consideration of the content, this is isomorphic to a boolean check, and so making that explicit communicates the intent more concisely. Moreover, a named function here also makes the comment on 203 unnecessary.

lib/analyse.ml
```diff
@@ -191,7 +195,14 @@ module Analysis = struct
             Current.Process.check_output ~cwd:dir ~cancellable:true ~job cmd >>= begin function
               | Error _ ->
                 (* new release *)
-                Lwt_result.return (add_pkg ~path ~name ~package New pkgs)
+                let package_dir = "packages/"^name in
+                let cmd = "", [| "git"; "show"; master^":"^package_dir |] in
+                Current.Process.check_output ~cwd:dir ~cancellable:true ~job cmd >>= begin function
```
shonfeder, line 200: Looks like we could get by with `exec` here, since we don't care about the output:

```suggestion
                Current.Process.exec ~cwd:dir ~cancellable:true ~job cmd >>= begin function
```

Not very important, but seems so save a fer characters, is a smaller change, and maybe saves an allocation. Feel free to ignore if you don't agree.

lib/analyse.ml
shonfeder, line 204: Since all we need to compute here is the kind of change, it think it could arguably be cleaner if this were something like (incorporating both my and Mark's previous suggestions):

```suggestion
                let+ is_newly_published = is_newly_published_package ~cwd ~job name master in
                let change = if is_newly_published then Package else Release in
                add_pkg ~path ~name ~package change_kind (New change)
```

This is arguably stylistic, so I wouldn't insist here. But I find these very long, deeply nested functions pretty hard to read compared to flatter alternatives, so I'm concerned with maintainability not just my fashion sense :)

shonfeder (commented):
Thank you for the cleanup here!

I think Mark's suggestion is a pretty good one, as it will simplify the case analysis and make it easier to read the logic at a glance, iiuc.

I also have a couple other suggestions. None are super critical, but the one I feel would help the most is using a named helper.

lib/analyse.ml
punchagan: This makes, sense. Thanks!

lib/analyse.ml
punchagan: Yes, I used `check_output` since we were using it for the check to detect new releases, but this makes sense. Thanks!

lib/analyse.ml
punchagan: We are currently using infix syntax for Lwt map/bind and `let+`, `let*`, etc for OCurrent. So, though stylistically I like your suggestion, this may be confusing. I've incorporated the change partially, still keeping the nested function.

punchagan:
Thanks for the review and suggestions, @mtelvers and @shonfeder. I've incorporated the changes and updated the PR. Let me know if there's anything else that needs improvement!

lib/analyse.ml
shonfeder: That makes sense!

shonfeder (approved):
Looks great!

PR #349, opened 2024-08-26
Remove submodules

mtelvers:
Alternative to https://github.com/ocurrent/opam-repo-ci/pull/320 aka https://github.com/tarides/infrastructure/issues/320


shonfeder:
To be sure I understand, the approach suggested here is to just depend on the published packages, and give up the ability to incorporate unpublished changes or to iterate quickly via changes to vendored deps?

shonfeder:
We talked about this yesterday and I am convinced this is a good idea. To summarize:

- These deps are now mature enough we don't need to be changing them very frequently.
- We currently are not incentivized to publish changes to the deps to opam, which means the community loses out on potential improvements, and moving to the published packages here will fix that.
- Forcing the infrastructure team to publish packages more regularly sets up a positive feedback loop to drive us to be more responsive to the needs of the opam repo.
- We will still have an escape hatch: if we do need to deploy a change involving a dependency and cannot wait for a release on opam, we can simply pin the package in question at the time, and supplying the hash for one package temporarily is much less cumbersome than what would have been needed by https://github.com/ocurrent/opam-repo-ci/pull/320.

Dockerfile.web
```diff
@@ -1,25 +1,9 @@
 FROM ocaml/opam:debian-12-ocaml-4.14@sha256:06d58a5cb1ab7875e8d94848102be43f6492e95320e8e9a9ecb9167654d0ee3f AS build
-RUN sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam && opam init --reinit -ni
+RUN sudo ln -f /usr/bin/opam-2.2 /usr/bin/opam && opam init --reinit -ni
```
shonfeder, line 2: I know it's a small change, but perhaps we should save this update for its own PR, so it'll be easier to isolate the cause if anything breaks between this update and the change in dependency management. WDYT, am I being paranoid?

Dockerfile
```diff
@@ -1,34 +1,15 @@
-FROM ocaml/opam:debian-12-ocaml-4.14@sha256:06d58a5cb1ab7875e8d94848102be43f6492e95320e8e9a9ecb9167654d0ee3f AS build
-RUN sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam && opam init --reinit -ni
+FROM ocaml/opam:debian-12-ocaml-4.14 AS build
```
shonfeder, line 1: Will is it possible we'll be inviting more instability by unpinning the base image? Or do you think that is not a concern?

An alternative is that we rely on published packages, but still pin our repository down via the image hash, requiring we update this as needed, but insuring packages updates don't break things out from under us... but then, maybe would catch that in our deployability checks (at least 99%)?

dune
```diff
@@ -1,3 +1,3 @@
 (dirs :standard \ var)
 
-(vendored_dirs ocurrent ocluster ocaml-dockerfile ocaml-version opam)
+(vendored_dirs opam)
```
shonfeder, line 3: I don't see an `opam` directory anywhere in the source tree. Can we get rid of this stanza entirely?

Dockerfile.web
mtelvers: I'm happy to leave this change until another time as it is unrelated to the other changes.

Dockerfile
mtelvers: We pin the base image SHA, and then checkout a specific opam-repository SHA.  Picking the base image SHA seems redundant as the latest image always includes the history.  By picking the latest image we reduce the work needed by `apt update`.  Therefore, I think the discussion should be on whether we fix the opam-repository SHA (line 4).  I took both out in the PR to raise the discussion.  As you say, we have a decision on whether we decide to move this forward each time we publish new packages on opam-repository, or just accept the current version.  On reflection, I think we should fix the opam-repository SHA as this will give us a reliably deployable package.  OCaml-CI will test with the latest opam-repository so we will get advanced warning on issues.  The PR to update the SHA will be the trigger to cause `ocurrent-deployer` to build and deploy the new version.

shonfeder: Your proposed way forward there makes sense to me. :+1:

punchagan:
> We talked about this yesterday and I am convinced this is a good idea. To summarize:

I agree with the summary presented here. 

I'm curious to know what the developer workflow would be like, though, when we want to update one of the dependencies (say `ocurrent`) to add a new feature? I like the convenience of changing the vendored dirs code and seamlessly having dune build the new changes, etc. Is there a way to replicate something like that even with this change?

shonfeder:
> I'm curious to know what the developer workflow would be like, though, when we want to update one of the dependencies (say `ocurrent`) to add a new feature?

This is a great question. I do think that this change will introduce some additional friction to that process, and this is just reflective of the fact that we are maturing to a looser coupling.

That said, there are two workflows I can think of that may help recover some of the benefits of the vendoring for development purposes:

1. Pinning the local dev repo, e.g., via `opam install --working-dir --yes ../ocurrent`. This makes it relatively easy to work on changes on a dependency locally and have those propagated here. This is simple to set up and only changes the opam dev switch. But the pin needs to be updated (i.e. by rerunning that command) after making changes, and that can slow down iterations a bit, as its quite a bit slower than an incremental build.

2. Using [opam-monorepo](https://github.com/tarides/opam-monorepo) could give us a smoother experience, at the cost of a bit of experimentation with the right configuration and workflows (e.g., committing lock dirs or not? Using `x-opam-monorepo-opam-provided` for most deps or not?).

Any other ideas?

Dockerfile.web
```diff
@@ -1,25 +1,9 @@
 FROM ocaml/opam:debian-12-ocaml-4.14@sha256:06d58a5cb1ab7875e8d94848102be43f6492e95320e8e9a9ecb9167654d0ee3f AS build
```
punchagan, line 1: We'd want to have this `Dockerfile.web` to be in sync with `Dockerfile`.

mtelvers: I have updated `Dockerfile.web`

shonfeder (approved):
LGTM! #355 will solve the ci errors, iiuc :)

PR #350, opened 2024-08-27
Use opam-ci-check lint tool in the linting step 

punchagan:
The code for the linting has been extracted to [ocurrent/opam-ci-check](https://github.com/ocurrent/opam-ci-check). This PR switches to using the `opam-ci-check` CLI in the linting step. 

We currently use the `opam-ci-check` repo as a submodule of `opam-repo-ci`, until we feel the project is ready to be released on `opam`, and to allow for faster development cycles when we need to fix/update `opam-ci-check`. This conflicts with the changes in #349. 

It would be nice to somehow have this PR tested on a bunch of opam-repository PRs before deploying it (in production). Not sure if we have a good way to do this, currently. 

Closes tarides/infrastructure#340


mtelvers:
For testing, you can setup PR's against mtelvers/opam-repository and these will be tested at http://5.102.169.187:8081.  Looks ok to me:

```
2024-08-27 09:04.29: New job: Lint
2024-08-27 09:04.29: Checking out commit 7e2b6041. To reproduce:
                       git clone --recursive "https://github.com/mtelvers/opam-repository.git" && cd "opam-repository" && git fetch origin "refs/pull/19/head" && git reset --hard 7e2b6041
2024-08-27 09:04.29: Exec: "cp" "-a" "--" "/home/mtelvers/opam-repo-ci/var/git/opam-repository.git-780a482f06f13c0e99a5fd9a91eca9badbd8accdecc0edc0362b986355279f43/.git" 
                           "/tmp/git-checkout2218085"
2024-08-27 09:04.30: Exec: "git" "-C" "/tmp/git-checkout2218085" "submodule" 
                           "deinit" "--force" "--all"
2024-08-27 09:04.30: Exec: "git" "-C" "/tmp/git-checkout2218085" "reset" 
                           "--hard" "-q" "7e2b6041950aa61d39f0f4b9bf79e1ea99c04cef"
2024-08-27 09:04.33: Exec: "git" "-c" "protocol.file.allow=always" "-C" 
                           "/tmp/git-checkout2218085" "submodule" "update" 
                           "--recursive" "--init" "--no-fetch"
2024-08-27 09:04.33: Exec: "git" "merge" "-q" "--" "ccefad276987c26cf93e400c096d40ab2d13c19e"
2024-08-27 09:04.33: Exec: "opam-ci-check" "lint" "-r" "." "-c" "ocaml-protoc-plugin.6.1.0"
Linting opam-repository at /tmp/git-checkout2218085/. ...
No errors
2024-08-27 09:04.46: Job succeeded
```

punchagan:
> For testing, you can setup PR's against mtelvers/opam-repository and these will be tested at https://opam-repo.tunbury.org.uk.

@mtelvers  I now seem to be getting a 502.  Not sure if something in this PR caused the service to go down.

mtelvers:
I've restored the test version.

shonfeder:
> We currently use the `opam-ci-check` repo as a submodule of `opam-repo-ci`, until we feel the project is ready to be released on `opam`, and to allow for faster development cycles when we need to fix/update `opam-ci-check`. This conflicts with the changes in #349.

Granted there will be a merge conflict, but I don't think the efforts necessarily conflict. Here are two possible ways forward that don't require us prematurely publishing `opam-ci-check`:

1. Remove those submodules, but add this submodules. I am OK with this, as the dependencies are in very different level of maturity.
2. Move this to a `pin-depends`, and just update pin hash as needed.

I think I am actually in favor of 2. It will help enforce the separation between the local first logic we want in `opam-ci-check` and will ensure no conflicts with #349.  WDYT?

lib/lint.ml
```diff
@@ -8,459 +8,33 @@ let () = Lwt_preemptive.init 0 1 (fun _errlog -> ()) (* NOTE: Lwt_preemptive is
...
+        | _ -> None
+      ) packages in
+    let new_ = List.filter_map (fun (pkg, change) ->
+        match change with
+        | Analyse.Analysis.(New Package) -> Some (OpamPackage.to_string pkg)
+        | _ -> None
+      ) packages in
+    let cmd = ["opam-ci-check"; "lint"; "-r"; "."] @
+              (if changed <> [] then ["-c"; String.concat "," changed ] else []) @
+              (if new_ <> [] then ["-n"; String.concat "," new_ ] else [])
```
shonfeder, line 35: I suggest moving the parsing of the `changed`/`new_` values into CLI flags be into the extraction step. IMO, it makes for a cleaner separation of concerns and makes the CLI invocation clearer.

This could be done in several ways, but would be something like:

```suggestion
    let changed =
      packages
      |> List.filter_map (fun (pkg, change) ->
          match change with
          | Analyse.Analysis.(New Release | Unavailable | SignificantlyChanged | InsignificantlyChanged ) -> Some (OpamPackage.to_string pkg)
          | _ -> None) 
      |> function
      | [] -> []
      | changed_packages -> ["--changed-packages"; String.concat "," changed_packages]
    in
    let new_ =
      packages
      |> List.filter_map (fun (pkg, change) ->
          match change with
          | Analyse.Analysis.(New Package) -> Some (OpamPackage.to_string pkg)
          | _ -> None) 
      |> function
      | [] -> []
      | new_packages -> ["--newly-published"; String.concat "," new_packages]
    in
    let cmd = ["opam-ci-check"; "lint"; "-r"; "."] @ changed @ new_
```

This way we only have to reason about what gets computed for these flags in one place.

lib/cluster_build.ml
```diff
@@ -17,7 +17,7 @@ let tail ?buffer ~job build_job =
     | Error (`Capnp e) -> Lwt.return @@ Fmt.error_msg "%a" Capnp_rpc.Error.pp e
     | Ok ("", _) -> Lwt_result.return ()
     | Ok (data, next) ->
-      Option.iter (fun b -> Buffer.add_string b data) buffer;
+      Stdlib.Option.iter (fun b -> Buffer.add_string b data) buffer;
```
shonfeder, line 20: I wonder if it might not be a better idea to disable implicit transitive dependencies in our build configuration: https://dune.readthedocs.io/en/stable/reference/dune-project/implicit_transitive_deps.html

I'm not crazy about the fact that adding a dependency for our CLI tool pollutes the global module name space for the entire project XD 

WDYT?

shonfeder (changes requested):
Awesome to see this dropping into place! :dancers:

punchagan:
> Move this to a `pin-depends`, and just update pin hash as needed.

I was mistakenly under the impression that we didn't want to use `pin-depends` since `ocaml-ci` is not happy with it. I just checked and clarified. I'll try updating this PR to use pin-depends, instead.

shonfeder:
Yeah, in depends can be used with ocaml-ci, only long with a hash, not a
branch name or left unqualified. My closed pr int his repo moving to pin
depends gives an example.

lib/cluster_build.ml
punchagan: Yes, your suggestion might be a better way to go about this. But, I ended up completely removing the commit. It is an ancient commit that was leftover from my experiments when I was trying to use the extracted code as a library, instead of a CLI tool.

lib/lint.ml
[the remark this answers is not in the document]
punchagan: Fair enough! Thanks for the suggestion.

punchagan: Makes sense. Changed!

punchagan:
The deployability check seems to fail because we "pinned" to much older version of `opam-repository` and `dune >= 3.15` isn't available.

shonfeder:
> The deployability check seems to fail because we "pinned" to much older version of `opam-repository` and `dune >= 3.15` isn't available.

Perhaps we could do a separate PR to update this pin of the repo and merge that in ahead of this one?

punchagan:
> Perhaps we could do a separate PR to update this pin of the repo and merge that in ahead of this one?

Sure. https://github.com/ocurrent/ocurrent-deployer/pull/225

punchagan:
Before we merge his PR, we might also want to bump up the hash for the pin-depends to the latest version.

punchagan:
Ah, I completely misunderstood how the deployability check was working. This makes, sense. Thanks!

lib/lint.ml
```diff
@@ -8,459 +8,40 @@ let () = Lwt_preemptive.init 0 1 (fun _errlog -> ()) (* NOTE: Lwt_preemptive is
...
-
-  let get_dune_project_version ~pkg url =
-    Lwt_io.with_temp_dir @@ fun dir ->
-    Lwt_preemptive.detach begin fun () ->
-      OpamProcess.Job.run @@
-      OpamRepository.pull_tree (OpamPackage.to_string pkg)
-        (OpamFilename.Dir.of_string dir)
-        (OpamFile.URL.checksum url)
-        [OpamFile.URL.url url]
-    end () >>= function
```
shonfeder, line 115: Great that we can remove this now! IIUC, this was needed because we were doing long-running operations in the processes by calling the opam libs. But now we are shelling that out, which means the program will not block on completion of the process, but can context switch while we wait for inputs.

We should also remove initialization of `Lwt_preemptive` on line 6 above.

shonfeder (changes requested):
Just one tiny fix to suggest. Looks great to me!

lib/lint.ml
punchagan: Thanks for catching this!

shonfeder (approved):
LGTM!

punchagan:
Merged. I'll attempt to deploy this now!

PR #351, opened 2024-09-04
Use 0install-solver in tests

shonfeder:
Meant to address persistent solvers timeouts causing CI tests to give false negatives.

Suggested by @kit-ty-kate

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -23,7 +23,7 @@ let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~revdep ~
    else
      []
   ) @
-  (if revdep || lower_bounds then
+  (if revdep || lower_bounds || with_tests then
      [
        run "opam option solver=builtin-0install";
      ]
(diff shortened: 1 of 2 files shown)


lib/opam_build.ml
```diff
@@ -23,7 +23,7 @@ let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~revdep ~
    else
      []
   ) @
-  (if revdep || lower_bounds then
```
shonfeder, line 26: With this change, we are using 0install in every case *except* `pin` I'm wondering if there's a good reason not to just set this unconditionally here. Thoughts?

kit-ty-kate: except also for the normal/build case which also use this function. The normal case should still use mccs whenever possible given it is the default solver and differences between solvers might mean one might fail but the other doesn't.
0install might be in a state to be the default in the near future but in the meantime i'm not sure it should be.
Also opam 2.0 does not have it, and opam 2.1 and 2.2 have an older version of it which is less close to what user encounter.

shonfeder: I see. Thanks.

shonfeder:
Motivated by discussion on #26473

punchagan (approved):
Thanks for the suggestion, @kit-ty-kate and the quick fix @shonfeder !

mtelvers (approved):

PR #355, opened 2024-09-05
Remove Cohttp.Response.Make

mtelvers:
`Cohttp.Response.Make` has been deprecated by https://github.com/mirage/ocaml-cohttp/pull/1024/files.

The module Response doesn't seem to be used so can we just drop it?


shonfeder (approved):

talex5 (approved):
Looks like it was actually deprecated by https://github.com/mirage/ocaml-cohttp/pull/806 (not one of mine!). But removing unused modules seems fine to me.

PR #357, opened 2024-09-17
Increase solver timeout

### LINKED ISSUE #354: Tests on opam versions =< 2.3 are failing with solver timeouts
A handful of tests fail with solver timeouts on many package PRs. See, e.g., https://github.com/ocaml/opam-repository/pull/26473#issuecomment-2328002771

Using opam 2.3 along with configuring opam to use the 0install solver (e.g., in https://github.com/ocurrent/opam-repo-ci/pull/343) has been a big a help in reducing solver timeouts, but the problem still persists, and it creates false positives in the repo CI and slows down package publication. It could also mask genuine test failures.

We still run tests on opam 2.0, 2.1, and 2.2, since those are available on some distros, so we cannot simply switch all tests over to using this approach (see https://github.com/ocurrent/opam-repo-ci/issues/352). An alternative approach proposed by @mtelvers in https://github.com/ocurrent/opam-repo-ci/pull/342 is to increase the timeout.

shonfeder:
Trying to fix #354.
Recreates #342 (for which the branch has been deleted).

If this timeout is long enough, it will address the timeouts on opam
2.0, 2.1, and 2.2, since this env var has been supported at least since 2.0:
https://github.com/ocaml/opam/blob/7d4a0f2e0fefe748efbd7358775b1a0bd8267544/src/solver/opamSolverConfig.ml#L131-L132

This can increase the build time of some tests, but I think it should be
worth it if it can give proper results:

- Only few builds (~3) per PR run on these older opam versions.
- We are wasting a lot of cumulative person time having to check the CI
  results for PRs that should be green if the CI wasn't giving these
  (arguably) false negatives.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -125,7 +125,7 @@ let setup_repository ?(local=false) ~variant ~for_docker ~opam_version () =
   run "uname -rs && opam exec -- ocaml -version && opam --version" ::
   env "OPAMDOWNLOADJOBS" "1" :: (* Try to avoid github spam detection *)
   env "OPAMERRLOGLEN" "0" :: (* Show the whole log if it fails *)
-  env "OPAMSOLVERTIMEOUT" "500" :: (* Increase timeout. Poor mccs is doing its best *)
+  env "OPAMSOLVERTIMEOUT" "1000" :: (* Increase timeout. Poor mccs is doing its best *)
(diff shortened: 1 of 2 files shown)


punchagan (approved):

mtelvers (approved):

shonfeder:
Thanks for the reviews!

PR #362, opened 2024-09-20
Fix lint job crash when no packages were changed

### LINKED ISSUE #361: Avoid running the lint step if no packages were changed
This [PR](https://github.com/ocaml/opam-repository/pull/26508), for instance, doesn't change any packages and the analysis correctly detects it. But, we still try to run a lint with no packages, and [fail](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/5850dbfbc9a7b287c7242765841029ed8469fc30/variant/(lint))

punchagan:
Closes #361

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -77,8 +77,13 @@ module Lint = struct
 
   let run { master } job { Key.src; packages } () =
     Current.Job.start job ~pool ~level:Current.Level.Harmless >>= fun () ->
-    Current_git.with_checkout ~job src @@ fun dir ->
-    Check.of_dir ~master ~job ~packages dir
+    match packages with
+    | [] ->
+      Current.Job.log job "No packages to lint.";
+      Lwt.return (Ok ())
+    | _ ->
+      Current_git.with_checkout ~job src @@ fun dir ->
+      Check.of_dir ~master ~job ~packages dir
   let pp f _ = Fmt.string f "Lint"
 
   let auto_cancel = true


lib/lint.ml
```diff
@@ -77,8 +77,13 @@ module Lint = struct
...
     Current.Job.start job ~pool ~level:Current.Level.Harmless >>= fun () ->
-    Current_git.with_checkout ~job src @@ fun dir ->
-    Check.of_dir ~master ~job ~packages dir
+    match List.length packages with
+    | 0 ->
+      Current.Job.log job "No packages to lint.";
+      Lwt.return (Ok ())
+    | _ ->
+      Current_git.with_checkout ~job src @@ fun dir ->
+      Check.of_dir ~master ~job ~packages dir
```
shonfeder, line 86: `List.length` needs to count all the elements of a list. Emptiness are checks are therefore better done by pattern matching, which take a single operation:

```suggestion
    match packages with
    | [] ->
      Current.Job.log job "No packages to lint.";
      Lwt.return (Ok ())
    | _ ->
      Current_git.with_checkout ~job src @@ fun dir ->
      Check.of_dir ~master ~job ~packages dir
```

This is also how [`List.is_empty`](https://github.com/ocaml/ocaml/blob/d753ea1992d27aa22dcb449f4a0ac178d167d52c/stdlib/list.ml#L571-L573) works.

shonfeder (approved):
The general approach here seems fine to me! I think is solution is nice, because we may end up wanting to run lint checks on thinks other than opam files.

shonfeder (changes requested):
Just the one list checking idiom request on my end!

lib/lint.ml
punchagan: Thanks for the mini-tutorial! Appreciate it. :)

shonfeder (approved):

Issue #359, opened 2024-09-18
Fix/improve lint reporting


shonfeder:
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/631a0a49c602befaee03f3e50eec470dca86eb8a

When the linting step fails it gives a report on the PR overview page that is hard sometimes quite hard to read, and, after the switch to using `opam-ci-check`, doesn't actually display the reason for the error.

E.g., we see that the CLI command failed, but do not see which errors triggered the failure until we look at the logs:

![2024-09-18-095258_1851x397_scrot](https://github.com/user-attachments/assets/d1ded07e-2039-4461-af09-0ee160036cb6)

I think the ideal output in this slot would be a bullet point list of the errors causing the linting step to fail.

PR #364, opened 2024-09-24
Fix lint reporting in the UI

punchagan:
image

The errors for the PR screenshot in ocurrent/opam-repo-ci#359 would look like this. :point_up:

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -13,6 +13,32 @@ module Check = struct
   let marshal () = Yojson.Safe.to_string `Null
   let unmarshal _ = ()
 
+  let parse_errors_from_log job =
+    let job_id = Current.Job.id job in
+    let path = Current.Job.log_path job_id in
+    match path with
+    | Ok path ->
+        Lwt_io.with_file ~mode:Lwt_io.input (Fpath.to_string path) (fun ch ->
+          let rec aux acc =
+            (* We choose the relevant output lines by filtering lines starting
+               with Error/Warning. The other log lines in the file would start
+               with a timestamp (logged info), or not start with the
+               Error/Warning prefix (other text output from the CLI tool). *)
+            Lwt_io.read_line_opt ch >>= function
+            | Some line ->
+              if (String.starts_with ~prefix:"Error" line ||
+                  String.starts_with ~prefix:"Warning" line)
+              then
+                aux (line :: acc)
+              else
+                aux acc
+            | None -> Lwt.return acc
+          in
+          aux [] >>= function
+          | [] -> Lwt.return (Error (`Msg "Could not find any Error/Warning lines in CLI output."))
+          | errors -> Lwt.return (Ok (String.concat "\n" errors)))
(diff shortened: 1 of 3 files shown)


lib/lint.ml
```diff
@@ -38,6 +65,18 @@ module Check = struct
...
     exec ~cwd ~job (cmd |> Array.of_list)
+    >>= function
+    | Error (`Msg err) ->
+      (* The exec function doesn't capture stdout when the command fails; so we
+         parse the command output from the log file, instead. *)
+        let errors = parse_errors_from_log job in
+        (match errors with
+        | Ok errors -> Lwt_result.fail (`Msg errors)
+        | Error (`Msg msg) ->
+          let error_msg = (Fmt.str "%s - %s" err msg) in
```
shonfeder, line 76: I think we will want to very clearly differentiate this case from errors that report a failing lint to users. Do you think this would help for that purpose?

```suggestion
          let error_msg = (Fmt.str "internal error: %s - %s" err msg) in
```

lib/lint.ml
```diff
@@ -13,6 +13,33 @@ module Check = struct
...
+                aux (line :: acc)
+              else
+                aux acc
+            | None -> acc
+          in
+          let errors = aux [] in
+          match errors with
+          | [] -> Error (`Msg "Could not find any Error/Warning lines in CLI output.")
+          | _ -> Ok (String.concat "\n" errors) )
+    | Error _ -> Error (`Msg (Fmt.str "Could not find the log file for job <%s>" job_id))
```
shonfeder, line 41: Should we try to preserve the err data given by `log_path` in case that is useful for debugging? E.g.,

```suggestion
    | Error (`Msg m) -> Error (`Msg (Fmt.str "Could not find the log file for job <%s>: %s" job_id m))
```

lib/lint.ml
```diff
@@ -13,6 +13,33 @@ module Check = struct
   let marshal () = Yojson.Safe.to_string `Null
   let unmarshal _ = ()
 
+  let parse_errors_from_log job =
+    let job_id = Current.Job.id job in
+    let path = Current.Job.log_path job_id in
+    match path with
+    | Ok path ->
+      In_channel.with_open_text (Fpath.to_string path) (fun ic ->
```
shonfeder, line 21: Sorry I didn't think of this before, but I'm wondering if we should be using `Lwt_io` functions here? I think these logs are usually pretty short, but in case they get a bit long, we may want to be sure that we will yield. WDYT?

shonfeder (commented):
Looking great! I had a few more thoughts I'd like your take on, but they are not major, and if you think I may be off the mark on any, happy to approve as is.

punchagan:
> Looking great! I had a few more thoughts I'd like your take on, but they are not major, and if you think I may be off the mark on any, happy to approve as is.

No, I appreciate your suggestions. I'll work on them on Monday.

shonfeder (approved):
LGTM! Thanks for this nice UI fix :tada:

PR #369, opened 2024-10-01
Add instructions to reproduce linting locally

### LINKED ISSUE #360: Provide local reproduction instructions for linting step
Now that we have the linting logic in a local-first tool, thanks to @punchagan's work culminating in https://github.com/ocurrent/opam-repo-ci/pull/350, we can provide users with instructions on how to run the tool locally.

I think printing this in the logs for the linting step probably makes the most sense for now? A good place to start may be with just a mock of how we want these instructions to look.

punchagan:
Closes #360

would conflict with #367. We can port the instructions to #367 directly, instead of merging this, if that seems better!

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -63,6 +63,16 @@ module Check = struct
           | new_ -> ["--newly-published"; String.concat "," new_ ]
     in
     let cmd = ["opam-ci-check"; "lint"; "--opam-repository"; "."] @ changed @ new_ in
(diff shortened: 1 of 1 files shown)


punchagan:
The instructions can be to pin to `live` branch once we merge #367

shonfeder (approved):
Looks great!

I've tested that the pin instruction works as expected from live.

PR #380, opened 2024-10-18
Some small cleanups and clarification

shonfeder:
I've been studying the code towards #403 and noticed some areas for small
improvements to simplify the data flow or clarify the types in question.

Review by commit should make the changes clear.


shonfeder:
I actually think this will benefit more from working thru the draft I have open, to arrive at a more holistic improvement here, rather than tweaking incremental stuff that may be redone immediately in followups.

PR #384, opened 2024-10-24
Expose the build-test step in the CLI for local runs

punchagan:
The PR is primarily a refactor to move the obuilder spec generation to `opam-ci-check`, and then exposing a CLI command to run the build/test step from the CLI. This should allow running the build/test step locally, wherever one can run docker.


opam-ci-check/lib/revdeps.ml
```diff
@@ -1,95 +0,0 @@
-(* SPDX-License-Identifier: Apache-2.0
```
shonfeder, line 1: Do we really want to rm all of this code at this point?

punchagan: Sure, we could leave it in. I removed it when our thinking was to use the obuilder-spec -> docker build route.

shonfeder:
The failing tests here should be addressed by #387. We should merge in the latter separately so that we confirm that these changes don't create any variation from the expected tests.

opam-ci-check/bin/main.ml
```diff
@@ -155,29 +125,34 @@ let lint_cmd =
...
-      const test_revdeps $ pkg_term $ local_opam_repo_term $ use_dune_term
-      $ no_transitive_revdeps)
+      const build_run_spec $ no_cache $ const BuildLowerBounds $ pkg_term
+      $ local_opam_repo_term)
     |> to_exit_code
   in
   let info =
-    Cmd.info "test" ~doc ~sdocs:"COMMON OPTIONS" ~exits:Cmd.Exit.defaults
+    Cmd.info "lower-bounds" ~doc ~sdocs:"COMMON OPTIONS"
+      ~exits:Cmd.Exit.defaults
```
shonfeder, line 155: Al alternative interface would give us a CLI like:

```
opam-ci-check build [--test-test] [--lower-bounds]
```

This would simplify the execution logic when constructing the build config, because the booleans from the flags could just be passed to `Spec.opam` directly, and the `build_spec` type would not be needed. It would also allow taking advantage of the ability to generate specs that have both flags set to true.

WDYT?

opam-ci-check/lib/lint.ml
```diff
@@ -172,8 +171,20 @@ module Checks = struct
...
+      |> Sexplib.Sexp.parse
+    with
+    | exception Sys_error _ -> Ok None
+    | Sexplib.Sexp.Done (List [ Atom "lang"; Atom "dune"; Atom version ], _) ->
+        Ok (Some version)
+    | Done _ -> Error "(lang dune ...) is not the first construct"
+    | Cont _ -> Error "Failed to parse the dune-project file"
+
+  let pull_pkg_get_version dir (pkg, url) =
+    let dir = Fpath.to_string dir in
```
shonfeder, line 187: Is this change related to the addition of the build-test functionality?

opam-ci-check/lib/test.ml
```diff
@@ -13,61 +12,45 @@ let error_to_string : error -> string =
...
-  let dirs =
-    H.create_dummy_projects parent opam_repository target selected_packages
+let build_run_spec ?(use_cache = true) ~opam_repository ~base config =
+  let spec = Opam_build.build_spec ~local:true ~for_docker:true ~base config in
+  let dockerfile =
+    Obuilder_spec.Docker.dockerfile_of_spec ~buildkit:false ~os:`Unix spec
   in
-  List.iter H.generate_lock_and_build dirs;
-  Ok ()
+  (* TODO: Maybe docker ability to build form git repos to default to official opam repo
```
shonfeder, line 23: These are by typos, sorry!

```suggestion
  (* TODO: Maybe use docker ability to build from git repos to default to official opam repo?
```

opam-ci-check/bin/main.ml
```diff
@@ -155,29 +125,34 @@ let lint_cmd =
...
+  Arg.value (Arg.flag info)
+
+let build_test_cmd =
+  let doc = "Build and test a package" in
   let term =
     Term.(
-      const show_revdeps $ pkg_term $ local_opam_repo_term
-      $ no_transitive_revdeps)
+      const build_run_spec $ no_cache $ const BuildTest $ pkg_term
+      $ local_opam_repo_term)
```
shonfeder, line 137: I think it would be helpful to have a CLI that allowed just dumping the obuilder spec (optionally translated to a docker file). Do you have any ideas on what a sensible CLI for that might look like?

shonfeder (commented):
Left a few questions and suggestions!

Two additional things to consider:

- What CLI should we design to let users specify the platform to use for the image used in the tests?
- How will we construct and print the CLI command for users to reproduce locally?

opam-ci-check/lib/lint.ml
punchagan: It's not. I was trying to remove some unnecessary helpers since we've already started using `Bos` in `opam-ci-check`. But, I'll send it as a different PR.

opam-ci-check/bin/main.ml
punchagan: I think it could be done as a flag to the `build` command to only dump the spec and not actually run the build step?

opam-ci-check/bin/main.ml
punchagan: This is a great suggestion. I've implemented it, thanks!

punchagan:
> * What CLI should we design to let users specify the platform to use for the image used in the tests?
> 
> * How will we construct and print the CLI command for users to reproduce locally?

As a starter, to allow using the same base images as the CI builds, I've added a WIP commit to allow using the docker base image tag as a CLI flag. 

The CLI instructions can become something like...

```
opam-ci-check build --with-test --base-image ocaml/opam:debian-12-ocaml-5.2@sha256:488fec4990a01d7738ad7fc4b2cb26228295f6364c32b834bfadaef79669d1f5  lwt.5.7.0
```

For other use cases, users can use variations of this like these:

```
opam-ci-check build --with-test --base-image ocaml/opam:debian-12-ocaml-5.2 lwt.5.7.0

opam-ci-check build --with-test --base-image debian-12-ocaml-5.2 lwt.5.7.0

opam-ci-check build --with-test --base-image ocaml-5.2 lwt.5.7.0

opam-ci-check build --with-test --base-image 5.2 lwt.5.7.0

opam-ci-check build --with-test --base-image debian-12 lwt.5.7.0

```

shonfeder (commented):
This is looking good! I think the CLI suggested here is heading in the right direction, but I think I spotted some possible problems.   

After thinking it over I came up with an alternate approach for consideration, and I've sketched it out in a branch. Looking forward to hearing what you think.

My last thought here is that it would be nice to have some testing for the build/test stuff. Ideally, imo, we'd do this by just printing obuilder specs (or dockerfiles) as cram tests. But I think we can pursue this as a followup.

opam-ci-check/bin/main.ml
```diff
@@ -70,6 +71,82 @@ let test_revdeps pkg local_repo_dir use_dune no_transitive_revdeps =
...
+  let distro = Stdlib.Option.value distro ~default:(default_distro ()) in
+  let compiler = Stdlib.Option.value compiler ~default:(default_compiler ()) in
+  (* TODO: Support other archs *)
+  let variant = Variant.v ~distro ~compiler ~arch:`X86_64 in
+  let base =
+    Spec.Docker
+      (if Option.is_some hash then Option.get base_image
+       else "ocaml/opam:" ^ Variant.docker_tag variant)
+  in
+  let config = Spec.opam ~variant ~lower_bounds ~with_tests ~opam_version pkg in
```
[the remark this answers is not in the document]
punchagan: Thanks for implementing the alternative, and also rebasing the branch against the current master! I've reset this branch to track that!

punchagan:
> After thinking it over I came up with an alternate approach for consideration, and I've sketched it out in a branch. Looking forward to hearing what you think.

This looks good to me. I started with the docker image hash as an argument since I was looking at the CI's hash and trying to provide a  way to use that. But, I like the separate arguments and better error handling of invalid arguments for the distro/compiler/etc to use. 


> My last thought here is that it would be nice to have some testing for the build/test stuff. Ideally, imo, we'd do this by just printing obuilder specs (or dockerfiles) as cram tests. But I think we can pursue this as a followup.

I've added some cram tests as suggested by you. Thanks!

shonfeder (approved):
I feel like we are in a great place here! Feel free to merger when you are happy with it or make any additional changes you think necessary.

opam-ci-check/bin/main.ml
```diff
@@ -218,20 +218,22 @@ let lower_bounds =
...
+let arch_options =
+  Ocaml_version.arches
+  |> List.map (fun arch ->
+      (Ocaml_version.string_of_arch arch, arch))
+
+let arch_conv = Arg.enum arch_options
+let distro_conv = Arg.enum distro_options
+
+(* TODO: This should be limited to an enum, but we can't currently compute  it
+   without depending on /lib/build.ml *)
```
punchagan, line 236: As a follow-up we could also use the [compiler_variants function](https://github.com/ocurrent/ocaml-version/blob/f92a9f1b848b5eed179861ea13586a618b5fa25b/ocaml_version.mli#L578) (needs `arch` and compiler varsion) to validate the options provided, before trying to fetch Docker images for the specified configuration.

Issue #328, opened 2024-06-26
Distinguish failing CI runs from CI errors


shonfeder:
A CI pipeline can fail because a build or test step returns a negative result, or because some error in the CI pipeline logic. We currently don't distinguish these outcomes in most cases. Ignoring this distinction has the following known down-sides:

1. Users will see job failures, and don't learn that the CI is suffering from internal errors until they inspect the logs.
2. While we collect metrics on the number of failed CI jobs, we cannot differentiate these from errors that would indicate sporadic or pervasive failures in the infrastructure and services.
3. We have no way of restarting jobs that failed due to an infrastructure error after that error has been repaired.

(2) bit us last week, when we failed to detect https://github.com/ocaml/infrastructure/issues/128 until it was so wide spread that users where noticing the failures. If we had monitoring that alerted us to errors in the pipeline, we could have seen this coming much earlier. Incorporating an `error` status into the metrics sent to Grafana would make these failures clearly visible.

To address (1), we can send an error status to GitHub. [The current API](https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#about-commit-statuses) supports the following statuses:

```
error, failure, pending, success
```

But, iiuc, we don't use the `error` status in our reporting:

https://github.com/ocurrent/opam-repo-ci/blob/97d42b7675b1de2400f167f9b13f5b04116cb541/service/github.ml#L21-L23

To address (2), we should send error results to Grafana:

https://github.com/ocurrent/opam-repo-ci/blob/97d42b7675b1de2400f167f9b13f5b04116cb541/service/metrics.ml#L34-L37

To address (3), we start by recording failures in the job index, which doesn't currently differentiate failures from errors:

https://github.com/ocurrent/opam-repo-ci/blob/97d42b7675b1de2400f167f9b13f5b04116cb541/lib/index.ml#L114

hannesm:
I propose to rename "error" to "internal error" or "internal CI error". And once this build status has been achieved, maybe an automatic restart of the job should be scheduled (surely with an exponential backoff to avoid CI going crazy)? Same could be done for "ocaml-ci".

PR #393, opened 2024-11-24
Adjust maintainer contact lint

shonfeder:
Closes https://github.com/ocaml/infrastructure/issues/152 by addressing
the subsequenst requests for tweaked behavior.

With this change in place, the lint will only fail if there is no email address
at all in the `maintainer` field AND no bug-tracker. We already issue a warning
when no bug-tracker is present, and since we currently make no distinction
between errors and warnings in our CI results, this check is made somewhat
redundant. However, once https://github.com/ocurrent/opam-repo-ci/issues/328 is
introduced, the distinction here will make a difference.

### DIFF
--- a/opam-ci-check/lib/lint.ml
+++ b/opam-ci-check/lib/lint.ml
@@ -265,13 +265,20 @@ module Checks = struct
         if is_build then (pkg, DuneIsBuild) :: errors else errors
     | None -> []
 
-  let check_maintainer_email ~pkg opam =
+
+  let check_maintainer_contact ~pkg opam =
+    let is_present bug_reports = bug_reports <> [] in
+    let includes_an_email maintainers =
+      List.exists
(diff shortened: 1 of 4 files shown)


opam-ci-check/lib/lint.ml
```diff
@@ -265,13 +265,20 @@ module Checks = struct
         if is_build then (pkg, DuneIsBuild) :: errors else errors
     | None -> []
 
+
   let check_maintainer_email ~pkg opam =
```
punchagan, line 269: nit: Maybe also rename the check function?

punchagan (approved):
Looks good to me modulo the CI failures and a minor nit pick. Thanks for taking care of this!

opam-ci-check/lib/lint.ml
shonfeder: Good catch. Thanks.

shonfeder:
Thanks for the review :)

PR #395, opened 2024-11-27
Refactor lint api

punchagan:
This PR is a preparatory refactor to separate out the Lint functionality from the project as a separate library for use with package publication tools to prevent badly linted PRs being opened on opam-repository.

The big API change is to use parsed OPAM.t objects instead of parsing the opam files in the lint library. This should make it easier to use the library API from the publication tools. This also allows us to run the lint before creating the package directory in the `opam-repository`. (In such cases, we skip the `check_package_dir` lint).

Also, the download of package sources in the `check_dune_constraints` test has been removed and the downloads are now moved to the `main` module. This allows us to prevent any package downloads as part of the library, and allows us to pass the path of the directory when using the API.

We also add some light-weight analysis to be able to infer if a package being linted is a "newly-published" package, or an update to an existing one. This makes the public API simpler. But, we also allow explicitly specifying this information to bypass the inference, if required. This can be used from the CLI by the CI.

The PR has a few minor clean-ups/improvements:

- Remove use of Option.default from extlib
- Minor formatting change with ocamlformat
- Ensure temp directory is always removed
- Rename `repo_dir` argument as `opam_repo_dir` for clarity


opam-ci-check/bin/main.ml
```diff
@@ -16,12 +18,68 @@ let to_exit_code : (unit, string) result Term.t -> Cmd.Exit.code Term.t =
...
+    Opam_helpers.path_from_pkg ~repo_dir:opam_repo_dir pkg // "opam"
+  in
+  (* NOTE: We use OpamFile.OPAM.read_from_channel instead of OpamFile.OPAM.file
+     to prevent the name and version fields being automatically added *)
+  In_channel.with_open_text opam_path (fun ic ->
+      let opam =
+        try Ok (OpamFile.OPAM.read_from_channel ic)
+        with OpamPp.Bad_format e | OpamPp.Bad_version (e, _) -> Error e
+      in
+      opam)
```
shonfeder, line 51: I find the `opam` variable a bit misleading here, because it may actually be an `Error _`, and so not have opam file data at all. IMO this would be clearer like

```suggestion
        try Ok (OpamFile.OPAM.read_from_channel ic)
        with OpamPp.Bad_format e | OpamPp.Bad_version (e, _) -> Error e)
```

or maybe

```suggestion
        try     
          let opam = OpamFile.OPAM.read_from_channel ic in
          Ok opam
        with 
          OpamPp.Bad_format e | OpamPp.Bad_version (e, _) -> Error e)
```

or maybe rename the `opam` variable to `result`?

opam-ci-check/lib/lint.mli
```diff
@@ -5,34 +5,55 @@
...
+(** [get_packages repo] is a list of all the packages in the opam repository
+    located at [repo].
+
+    Example:
+
+    {[
+      let all_packages = get_packages repo
+    ]} *)
+
+val make_parse_error : OpamPackage.t -> OpamPackage.t * error
```
shonfeder, line 26: With the new API, which expects the opam file to already be parsed, I don't think it makes sense for the parse error to be part of the lint API... so, IMO, this suggest a choice:

1. Either accept that parsing opam files is out of scope for the CI linting logic, and move that into (e.g.) the analysis stage.
2. Keep parsing as part of the the linting.

Support for (1) may be drawn from considering that we are already parsing all opam files ar part of the analysis phase (https://github.com/ocurrent/opam-repo-ci/blob/ec9e8ae658a0b6363db848db4b1d8d0082db034d/lib/analyse.ml#L142-L148), and noting that https://github.com/ocaml/opam-repository/issues/26956 indicates we may want opam parsing failures to be more visible category of failure than just a linting error.

Support for (2) may be found in the fact that parsing errors have been parting of linting up to this point, and that we generally expect parse failures to be reported in our linting tools.

I suggest we follow (2), even if we ultimately want to consider (1), because it is a smaller change. To achieve (2), we could tweak the new API so that `lint_package` takes a `string` as its principle argument, and then parse the string internally. If we want to support both a string and an already parsed opam file, we could have the principle argument be something like

```
[`Parsed of OpamFile.OPAM.t | `Raw of string]
```

It is possible I am being to pedantic, but with long term viability in view, I think it is not a great idea to have a package like this holding a data type for a kind of data it cannot produce.

WDYT?

shonfeder (commented):
I think this will make for a quite nice API change! I had several questions and thoughts as I read thru, many of them are likely just design decisions to work out.

opam-ci-check/bin/main.ml
```diff
@@ -16,12 +18,68 @@ let to_exit_code : (unit, string) result Term.t -> Cmd.Exit.code Term.t =
...
+         [ OpamFile.URL.url url ]
+  in
+  let dir =
+    match res with
+    | OpamTypes.Not_available (_, msg) ->
+        print_endline msg;
+        None
+    | Up_to_date _ | Result _ -> Some dir
+  in
+  f dir
```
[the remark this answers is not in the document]
punchagan: Yes, I did go with the implementation you suggested before, but I wanted to keep the old behavior through this API change. Don't entirely skip running lint checks if the package source couldn't be downloaded. Only the dune constraints check was being skipped. 

We could discuss the pros and cons and decide to change the behavior if we think completely skipping lints is better in those cases. Maybe it's a way to validate that the source URL is a valid one, even in CI? And we just re-run the lint in case of network errors?

opam-ci-check/lib/lint.ml
```diff
@@ -313,7 +306,14 @@ module Checks = struct
...
     in
-    get_files dir |> List.map check_file |> List.concat
+    (* FIXME: Would it be better to make skipping this check more explicit? *)
+    if Sys.file_exists dir then
+      get_files dir |> List.map check_file |> List.concat
+    else (
+      print_endline
+      @@ Printf.sprintf
+           "Skipped check_package_dir since package dir %s doesn't exist" dir;
+      [])
```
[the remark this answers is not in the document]
punchagan: The publication tools first generate the opam metadata (OpamFile.OPAM.t) objects, and I was trying to integrate the lints at that stage, rather than after they create the required files.

opam-ci-check/lib/lint.ml
```diff
@@ -263,7 +250,13 @@ module Checks = struct
...
         in
         if is_build then (pkg, DuneIsBuild) :: errors else errors
-    | None -> []
+    | None ->
+        print_endline
+        @@ Printf.sprintf
+             "Skipping check_dune_constraints since package dir for %s not \
+              provided"
+             (OpamPackage.to_string pkg);
+        []
```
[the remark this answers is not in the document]
punchagan: I was trying to keep the old behavior (in CI) where we run all the other lints, other than the dune constraints check, which would be skipped if we failed to download the package sources. We could discuss if that's desirable behavior or not.

opam-ci-check/lib/lint.mli
punchagan: I agree that the parse_error no longer should be a part of the API. And we can rely on the analysis phase catching any broken opam files.  Currently, we shouldn't even be triggering the lint step in cases where the analysis fails. 

For, the CLI also we can fail hard instead of trying to continue when one of the files fails to parse. I don't see a use case for when we'd want to allow one package with a parse error while parsing other files. I think it will also tie in with the other suggestion to remove `get_packages` from the public API and provide a bulk processing API.

opam-ci-check/lib/lint.mli
```diff
@@ -5,34 +5,55 @@
...
     - Case-insensitive string equality considering underscores ([_])
       dashes ([-]), and the empty string to be equal *)
 end
 
-val check :
-  new_pkgs:string list ->
-  changed_pkgs:string list ->
-  string ->
+val get_packages : string -> string list
+(** [get_packages repo] is a list of all the packages in the opam repository
```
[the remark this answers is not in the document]
punchagan: I was using it in the publication tools to list existing packages in the opam repository to find out if a package is new or not. 

There's a slight difference in when we are calling the lint using the CLI vs using the API. In the CLI, we use files already added to the opam-repository packages, while in the publication tools we run the lint before we have added the files for the current package to the opam-repository.  In the latter case, we can use `get_package` as a hackish analysis step to identify newly published packages. 

But, maybe the right way here would be to actually do a light weight version of the analysis in the library, instead?

opam-ci-check/lib/lint.ml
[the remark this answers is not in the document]
shonfeder: I agree it makes sense to keep the current logic for now. We can follow up with a change to this in the future. Thanks for the sync and ansync explanations :)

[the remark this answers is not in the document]
punchagan: We now return an error when the package src dir is not provided. The pkg_src_dir is still an option type to make it easy to run other checks, and only fail the check where the pkg_src_dir is used.

opam-ci-check/lib/lint.mli
punchagan: I've refactored this to completely get rid of the ParseError since we don't need to be "well behaved" when an opam file fails to parse, and can fail hard.

punchagan:
The inference of "newly published" or not isn't yet exposed on the CLI, and we always need to explicitly specify that from the CLI, as of now. But, I think the rest of the API change is self contained and useful to merge even without the CLI improvement.

punchagan:
> The inference of "newly published" or not isn't yet exposed on the CLI, and we always need to explicitly specify that from the CLI, as of now. But, I think the rest of the API change is self contained and useful to merge even without the CLI improvement.

I pushed a WIP commit to expose the newly-published inference and also to allow specifying the pkg_src_dirs from the CLI. I've not yet changed the usage of the CLI in the CI to use this CLI. This needs to be done once we are happy with the interface.

opam-ci-check/lib/lint.ml
```diff
@@ -367,40 +362,48 @@ module Checks = struct
...
+
+(** A package is considered newly published if the repository doesn't have any
+    other versions of the package, other than the current one.*)
+let is_newly_published ~opam_repo_dir pkg =
+  let pkg_name = OpamPackage.(pkg |> name |> Name.to_string) in
+  let pkg_str = OpamPackage.to_string pkg in
+  match get_files (opam_repo_dir // "packages" // pkg_name) with
+  | exception Sys_error _ -> true
+  | [] -> true
+  | v :: [] when v = pkg_str -> true
```
shonfeder, line 390: I find this a bit a cleaner, but probably just a matter of taste:

```suggestion
  | [v] -> v = pkg_str
```

opam-ci-check/bin/main.ml
```diff
@@ -140,35 +194,63 @@ let pkg_term =
...
+let package_specs_term =
+  let package_spec_conv =
+    let parse_package_spec s =
+      try
+        let parts = String.split_on_char ':' s in
+        let name_version = List.hd parts in
+        let pkg = OpamPackage.of_string name_version in
+        let attrs =
+          match List.tl parts |> String.concat ":" with
+          | "" -> []
```
shonfeder, line 206: IIUC, given the docs on the spec format, isn't this only valid if the split list is either a singletion or a two item list? It dosen't seem like more than one `:` should be able to occur in a term. So I think this maybe could be given better error handling with something like

```suggestion
        let parts = String.split_on_char ':' s in
        let* name_version, attrs = match parts with
          | [n] -> Ok (n, [])
          | [n; specs] -> Ok (n, String.split_on_char ',' specs)
          | _ -> Error ...
        in
        let pkg = OpamPackage.of_string name_version in
        let attrs = (* parsing apprs, with error handling, etc. *)
```

opam-ci-check/test/lint.t
```diff
@@ -57,44 +59,49 @@ Test the following:
 - [b.0.0.6] has a incorrectly formatted opam file
 - [b.0.0.7] has opam lint errors/warnings
 
-  $ opam-ci-check lint -r . -c b.0.0.1
+  $ opam-ci-check lint -r . b.0.0.1:new=false
```
shonfeder, line 62: Is there test to cover the automatic detection of whether package is new?

opam-ci-check/bin/main.ml
```diff
@@ -140,35 +194,63 @@ let pkg_term =
...
-      `Error
-        ( false,
-          "You must provide at least one changed or newly published package." )
-    else `Ok (changed_pkgs, newly_published_pkgs)
+    Arg.info []
+      ~doc:
+        "List of package specifications (format: \
+         <name.version>[:src=<path>][,new=<true|false>]). If [src] is not \
+         specified, the sources are downloaded from the source URL. If [new] \
+         is not specified, it is inferred from the opam repository."
```
shonfeder, line 245: This seems like it will solve the problem we are currently facing in CLI flexibility, but I am also a bit nervous about introducing a new ad hoc specification format.

If we need to provide rich, structured data like this, an alternative might be to just use s-expressions (we already depend on Sexplib any how :smile:). E.g.,

```
opam-ci-check lint p.0.1.0 '(q.0.1.1 (src some/dir))' '(r.1.0 new)' '(s.1.1 (src a/dir) new)'
```

I don't think this is preferable, but the benefit IMO is that doesn't require inventing new format for compound data. But I am also don't want to waste time bikeshedding on a CLI input format that likely will not be used much by people (because of the upstreaming plans).

Perhaps this worth a short sync chat on the tradeoffs?

shonfeder (changes requested):
Generally looking good to me, and offers a sensible solution to the CLI issue where we discussing. As I noted here, and in our previous discussion, I am uneasy about introducing new compound data formats, but I am not is worth blocking anything based in that unease :sweat_smile: 

I'm seeing the CI tests failing, and there are at least two points where I think clarification of improvements should be considered here before merging, regardless.

punchagan:
> I'm seeing the CI tests failing, and there are at least two points where I think clarification of improvements should be considered here before merging, regardless.

Yes, I didn't spend time on fixing all the CI tests because I wanted to see if we like the interface, first. Having a quick sync chat to decide this, sounds fine to me!  I'm also happy to open the CLI changes as a separate PR if that makes it easier to merge the other changes.

opam-ci-check/test/lint.t
punchagan: The name collision test now also tests the inference

opam-ci-check/bin/main.ml
```diff
@@ -140,35 +194,63 @@ let pkg_term =
...
+        in
+        (* Allow only named attributes *)
+        List.iter (fun x -> assert (String.contains x '=')) attrs;
+        let src =
+          List.find_opt (fun x -> String.starts_with ~prefix:"src=" x) attrs
+          |> Option.map (fun x -> String.sub x 4 (String.length x - 4))
+        in
+        let newly_published =
+          List.find_opt (fun x -> String.starts_with ~prefix:"new=" x) attrs
+          |> Option.map (fun x -> String.sub x 4 (String.length x - 4) = "true")
```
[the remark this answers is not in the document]
punchagan: Thanks for improving this!

opam-ci-check/lib/lint.ml
```diff
@@ -367,40 +362,51 @@ module Checks = struct
...
+}
+
+let v ~pkg ?(newly_published = None) ~pkg_src_dir opam =
+  { pkg; newly_published; pkg_src_dir; opam }
+
+(** A package is considered newly published if the repository doesn't have any
+    other versions of the package, other than the current one.
+
+    NOTE: If two versions of a package are published in the same PR, this
+    inference would fail to detect the package as new.*)
```
shonfeder, line 386: This is a very helpful pointer! :+1:

shonfeder (approved):
Looks great! Thanks for all the careful work on this, and for ensuring we prioritize trying to meet needs of package authors and repo maintainers!

opam-ci-check/lib/lint.ml
punchagan: It's possible to improve the inference in the cases where all the changed versions of a package have been passed to `lint`, but I'm not sure if it's worth fixing things for that edge case, right now.

punchagan:
> Looks great! Thanks for all the careful work on this, and for ensuring we prioritize trying to meet needs of package authors and repo maintainers!

Thanks for all your reviews, suggestions and brainstorming! It was fun. :)

PR #397, opened 2024-11-29
Don't fail analysis on new directories in the root

shonfeder:
The current analysis logic is overly restrictive in processing the directory structure of the repository: it only allows for the subdirectories `.github` and `packages`. The addition of any other subdirectory will result in an error like 

```
Job failed: Unexpected path "some/path" in output (expecting 'packages/name/pkg/...')
```

This is causing the analysis to fail on https://github.com/ocaml/opam-repository/pull/26992.

The view taken in this change is that the addition of files in new subdirectories other than `packages/` should be reviewed as normal PRs to the
code base.

A more restrictive view could be taken  that requires any additional directories be white-listed in advance. I am open to entertaining that view, and adopting that approach instead, if needed, but I don't see a good reason to be so restrictive currently.

cc @mseri and @hannesm

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -195,8 +195,6 @@ module Analysis = struct
         | Error _ as err -> Lwt.return err
         | Ok pkgs ->
           match String.split_on_char '/' path with
-          | [_] | ".github"::_ ->
-            Lwt_result.return pkgs
           | "packages" :: name :: package :: "files" :: _ ->
             Lwt_result.return (add_pkg ~path ~name ~package SignificantlyChanged pkgs)
           | ["packages"; name; package; "opam"] ->
@@ -211,8 +209,11 @@ module Analysis = struct
                 get_opam ~cwd:dir path
                 >>= add_changed_pkg ~path ~name ~package ~old_content pkgs
             end
+          | "packages" :: _ ->
+            Fmt.failwith "Unexpected path %S in packages directory (expecting 'packages/name/name.l.m.n/{opam,files/*}')" path
           | _ ->
-            Fmt.failwith "Unexpected path %S in output (expecting 'packages/name/pkg/...')" path
+            (* Changes to any other path should be reviewed manually, since it's a change into repository meta data *)
+            Lwt_result.return pkgs
       ) (Ok OpamPackage.Map.empty)
 
   let has_tests opam =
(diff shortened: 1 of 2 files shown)


shonfeder:
The CI failures are from a breaking change to Cohttp library, fixed in https://github.com/ocurrent/opam-repo-ci/pull/398, which we should merge in ahead of this one.

lib/analyse.ml
```diff
@@ -211,8 +209,11 @@ module Analysis = struct
                 get_opam ~cwd:dir path
                 >>= add_changed_pkg ~path ~name ~package ~old_content pkgs
             end
+          | "packages" :: _ ->
+            Fmt.failwith "Unexpected path %S in packages directory (expecting 'packages/name/name.l.m.n/{opam,files/*}')" path
```
punchagan, line 213: This could be changed to disallow even the `files` directory since https://github.com/ocurrent/opam-repo-ci/pull/336. We currently expect only the `opam` file and no other files.

test/analyse.t
```diff
@@ -111,6 +111,15 @@ See https://github.com/ocurrent/opam-repo-ci/issues/291
   $ opam-repo-ci-local --repo="." --branch=analyze-invalid-opam-file --analyse-only --no-web-server
   Error ""packages/a-1/a-1.0.0.1/opam" failed to be parsed: '.' is not a valid token"
 
-Clean up the build cache
+  $ rm -rf var
 
-  $ rm -rf ./var
+Test that a new root directory can be created.
```
punchagan, line 116: Thanks for the test!

punchagan (approved):
Looks good, other than a suggestion that could also be a follow-up. Thanks!

lib/analyse.ml
shonfeder: True, but I think the linting error you added is a nicer UI feedback than this one:

https://github.com/ocurrent/opam-repo-ci/blob/6f874c2bfd0cfc8ba1f9d59cf594c445216edaee/opam-ci-check/lib/lint_error.ml#L94-L98

Maybe we can consider a consolidation on a followup tho?

shonfeder:
Thanks for the review! I also discussed this in the opam-repistory-maintainers discord channel, and there was agreement it makes sense to allow this and no concerns voiced. I will merge and deploy.

PR #398, opened 2024-11-29
Update for cohttp 6 Response API

shonfeder:
Cohttp 6 introduced a breaking change to the `Response` API dropping
`flush` in https://github.com/mirage/ocaml-cohttp/pull/1052

This updates our usage to be compatible.


punchagan (approved):
Thanks for fixing this!

Issue #352, opened 2024-09-05
Regression from use of `opam ---option` flag in tests


shonfeder:
Introduced by #351 

```
/home/opam: (run (shell "opam option solver=builtin-0install"))
opam: unknown command `option'.
Usage: opam COMMAND ...
Try `opam --help' for more information.
"/usr/bin/env" "bash" "-c" "opam option solver=builtin-0install" failed with exit status 2
2024-09-05 11:20.39: Job failed: Failed: Build failed
```

https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/8bb0696dddd93b63ad5a6ad3c5047beaa2d3bacf/variant/extras,opam-2.0-ocaml-4.14,gluten-lwt.0.5.2,tests

mtelvers:
It's opam 2.0.10, so we need to update that to 2.2, which will help the timeout issue as well!
```
/home/opam: (run (shell "uname -rs && opam exec -- ocaml -version && opam --version"))
Linux 5.15.0-119-generic
The OCaml toplevel, version 4.14.2
2.0.10
```

shonfeder:
Doing so! Thanks :)

shonfeder:
I'm finding the code that sets the opam version quite tricky to follow. As a followup, I think this is probably worth a deep refactoring pass.

It *looks* like we are setting the opam version for all the obuilder specs at

https://github.com/shonfeder/opam-repo-ci/blob/e682604cd3f890291107f724d05e9e3896ed5be1/lib/build.ml#L9

But obviously something else is overriding this at some point, or we have another source of truth for which opam to use that only comes into effect on some platforms.

shonfeder:
Ah, ok, it is coming from

https://github.com/ocurrent/opam-repo-ci/blob/e682604cd3f890291107f724d05e9e3896ed5be1/lib/build.ml#L156-L159

And the reason this is testing on opam < 2.2 is because these tests are meant to be testing the lower opam versions XD

shonfeder:
So, iiuc, the only place where not already using the dev version is the place where we explicitly wanted to be testing older opam versions.

mtelvers:
The alternative to `opam option solver=builtin-0install` is to use `env "OPAMEXTERNALSOLVER" "builtin-0install";` but that also gave problems.  I agree, let's revert until we have a better plan.

shonfeder:
[...]

PR #399, opened 2024-12-02
Use 0install solver

mtelvers:
We continue to see solver timeouts as reported by @mseri and collated [here](https://github.com/mtelvers/solver-timeouts).  This PR changes the default solver from mccs to 0install, which is the plan for future versions of opam.

I initially tried increasing the timeout again from 1000 to 1500, but even this was not long enough in some cases.

I have previously argued against changing the solver as this would make the CI differ from the user experience. However, users must already be mitigating this, as I do not expect they are waiting more than 25 minutes for a solution. The default timeout is 60 seconds!

Since opam-repo-ci is testing the instability of the package, maintainers will likely be ignoring cases where the solver times out.  Therefore, it would be more useful to get results than fail on the timeout.  Furthermore, this should speed things along as the 0install is much faster.

### DIFF
--- a/opam-ci-check/lib/opam_build.ml
+++ b/opam-ci-check/lib/opam_build.ml
@@ -9,7 +9,7 @@ let cache ~variant =
                 Obuilder_spec.Cache.v "homebrew" ~target:"/Users/mac1000/Library/Caches/Homebrew" ]
 let network = ["host"]
 
-let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~revdep ~pkg =
+let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~pkg =
   let pkg_s = OpamPackage.to_string pkg in
   let with_tests_opt = if with_tests then " --with-test" else "" in
   let cache = cache ~variant in
@@ -19,13 +19,6 @@ let opam_install ~variant ~opam_version ~pin ~lower_bounds ~with_tests ~revdep ~
        env "OPAMCRITERIA"        "+removed,+count[version-lag,solution]";
        env "OPAMFIXUPCRITERIA"   "+removed,+count[version-lag,solution]";
        env "OPAMUPGRADECRITERIA" "+removed,+count[version-lag,solution]";
-     ]
-   else
-     []
-  ) @
-  (if revdep || lower_bounds then
-     [
-       run "opam option solver=builtin-0install";
      ]
    else
      []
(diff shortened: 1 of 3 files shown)


opam-ci-check/lib/opam_build.ml
```diff
@@ -122,10 +115,9 @@ let setup_repository ?(local=false) ~variant ~for_docker ~opam_version () =
      Otherwise the "opam pin" after the "opam repository set-url" will fail (cannot find the new package for some reason) *)
   run "%s -f %s/bin/opam-%s %s/bin/opam" ln prefix opam_version_str prefix ::
   run ~network "opam init --reinit%s -ni" opamrc :: (* TODO: Remove ~network when https://github.com/ocurrent/ocaml-dockerfile/pull/132 is merged *)
-  run "uname -rs && opam exec -- ocaml -version && opam --version" ::
+  run "opam option solver=builtin-0install && opam config report" ::
```
shonfeder, line 118: We tried enabling this in tests before, in https://github.com/ocurrent/opam-repo-ci/pull/351, which produced the regression in https://github.com/ocurrent/opam-repo-ci/issues/352

The problem was with the tests with earlier opam versions we are testing in https://github.com/ocurrent/opam-repo-ci/blob/e682604cd3f890291107f724d05e9e3896ed5be1/lib/build.ml#L156-L159

The conclusion we reached at that time was https://github.com/ocurrent/opam-repo-ci/issues/352#issuecomment-2332085848 . There may be some other useful context. See also https://github.com/ocurrent/opam-repo-ci/issues/354 .

AFACT, there is nothing in in this change that will avoid the problems with hit with opam 2.2, or is there and I am missing it?

So this makes me wonder if the solver timeouts may be a reason for us to try to accelerate https://github.com/ocurrent/docker-base-images/issues/132 and dropping support for opam 2.0 and 2.1 here...

I know we have wanted to keep testing on opam versions so long as they are in the repos of maintained distros, but if the quality of the tests are degrading across the board as a result, I think it may be counter productive. This just follows your compelling logic here, I think.

WDYT?

opam-ci-check/lib/opam_build.ml
```diff
@@ -122,10 +115,9 @@ let setup_repository ?(local=false) ~variant ~for_docker ~opam_version () =
      Otherwise the "opam pin" after the "opam repository set-url" will fail (cannot find the new package for some reason) *)
   run "%s -f %s/bin/opam-%s %s/bin/opam" ln prefix opam_version_str prefix ::
   run ~network "opam init --reinit%s -ni" opamrc :: (* TODO: Remove ~network when https://github.com/ocurrent/ocaml-dockerfile/pull/132 is merged *)
-  run "uname -rs && opam exec -- ocaml -version && opam --version" ::
+  run "%sopam config report" (match opam_version with `V2_1 | `V2_2 | `V2_3 | `Dev -> "opam option solver=builtin-0install && " | `V2_0 -> "") ::
```
shonfeder, line 118: I'm assuming this implies you've tested to confirm that this works on opam 2.1!

shonfeder (approved):
I think that accounts for my one reservation. Glad we are making this move!

opam-ci-check/lib/opam_build.ml
mtelvers: Yes, opam 2.1 onwards supports `opam option`, and 0install is already [enabled](https://github.com/ocurrent/ocaml-dockerfile/blob/dc5ab758fd581ac55b340af6060a7552d50200ba/src-opam/opam.ml#L242) at compile time.

PR #402, opened 2024-12-16
Skip deleted packages in analysis

punchagan:
Potential fix for comment [here](https://github.com/ocaml/opam-repository/pull/27100#issuecomment-2541870979).

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -245,9 +245,12 @@ module Analysis = struct
     Fpath.v "packages" // name // (name^"."^version) // "opam"
 
   let add_package_data ~dir (pkg, kind) packages =
-    match packages with
-    | Error _ as err -> Lwt.return err
-    | Ok packages ->
+    match (packages, kind) with
+    | Error _ as err, _ ->
+      Lwt.return err
+    | Ok packages, Deleted ->
+      Lwt.return_ok ((pkg, {kind; has_tests = false}) :: packages)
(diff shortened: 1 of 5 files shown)


lib/analyse.ml
```diff
@@ -244,10 +244,14 @@ module Analysis = struct
...
-    match packages with
-    | Error _ as err -> Lwt.return err
-    | Ok packages ->
+  let add_package_data ~dir ~job (pkg, kind) packages =
+    match (packages, kind) with
+    | Error _ as err, _ ->
+      Lwt.return err
+    | Ok packages, Deleted ->
+      Current.Job.log job "Skipping deleted package: %s" (OpamPackage.to_string pkg);
+      Lwt.return_ok packages
```
shonfeder, line 253: I think this fix makes sense at the analysis stage. Crashing with an internal error on this case is not good behavior!

shonfeder (commented):
I think this fix makes good sense! Thank you for helping ensure the archiving effort can proceed smoothly.

Do you think you'd be able to a test for this cas without too much trouble?

However, outside of the archiving process, I think we do not ever want packages deleted, so I wonder if we should make an issue to add some logic to the linting that would raise a warning for deleted packages? WDYT?

punchagan:
> Do you think you'd be able to a test for this cas without too much trouble?

I've added a cram test for this case.

shonfeder (approved):

PR #403, opened 2024-12-16
Switch to using opam-ci-check to generate the list of revdeps

punchagan:
Closes https://github.com/tarides/infrastructure/issues/422

This PR switches to using opam-ci-check to list reverse dependencies from the 3 command opam list invocation we previously had. The output is slightly different from the previous output in that it contains only available packages instead of all packages.  The PR also changes `opam-ci-check` CLI to accept an additional `use-default-root` argument to prevent creating a new opam root for running it's commands.

### DIFF
--- a/test/specs.expected
+++ b/test/specs.expected
@@ -1106,7 +1106,8 @@ list-revdeps: debian-12-ocaml-4.14/amd64 opam-dev
     COPY --chown=1000:1000 . opam-repository/
     RUN opam repository set-url --strict default opam-repository/
     RUN opam update --depexts || true
-    RUN echo '@@@OUTPUT' && opam list -s --color=never --depends-on 'a.0.0.1' --coinstallable-with 'a.0.0.1' --all-versions --depopts && opam list -s --color=never --depends-on 'a.0.0.1' --coinstallable-with 'a.0.0.1' --all-versions --recursive && opam list -s --color=never --depends-on 'a.0.0.1' --coinstallable-with 'a.0.0.1' --all-versions --with-test --depopts && echo '@@@OUTPUT'
+    RUN opam pin add -k git -y opam-ci-check git+https://github.com/ocurrent/opam-repo-ci.git#live
+    RUN echo '@@@OUTPUT' && opam exec -- opam-ci-check list --use-default-root 'a.0.0.1' && echo '@@@OUTPUT'
 
 build: debian-12-ocaml-4.14/amd64 opam-dev with-tests
     FROM BASE_IMAGE_TAG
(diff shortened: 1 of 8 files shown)


opam-ci-check/bin/main.ml
```diff
@@ -186,6 +187,13 @@ let no_transitive_revdeps =
   in
   Arg.value (Arg.flag info)
 
+let use_default_root =
+  let info =
+    Arg.info [ "use-default-root" ]
+      ~doc:"Use the default opam root in ~/.opam instead of creating a new one."
+  in
+  Arg.value (Arg.flag info)
+
```
shonfeder, line 196: Would it be possible to do this in a separate, preparatory PR?

This is not a huge changeset, but it will impact the whole pipeline, so if we can isolate the pat that will impact the CI as much as possible, that would probably be prudent.

shonfeder (changes requested):
Very happy to see this dropping in place to end out the year. Great work! :tada: 

Just one preliminary request on the PR. Let me know if if there is a reason why breaking out the default default root change into its own PR isn't feasible!

punchagan:
> Just one preliminary request on the PR. Let me know if if there is a reason why breaking out the default default root change into its own PR isn't feasible!

I've pulled out the changes to allow using the default opam root into #404. This PR can be reviewed once that has been merged.

shonfeder (changes requested):
This change looks great. Thanks for the tiny diff!

In the tests (which need to be updated), we now lose visibility of what revdeps are going to be tested. Do you think it would be easy enough to rig up a cram test for the the revdeps listing of opam-ci-check to go along with this change, so we can ensure we are inferring the revdeps we think we should?

test/specs.expected
```diff
@@ -1106,7 +1106,8 @@ list-revdeps: debian-12-ocaml-4.14/amd64 opam-dev
     COPY --chown=1000:1000 . opam-repository/
     RUN opam repository set-url --strict default opam-repository/
     RUN opam update --depexts || true
-    RUN echo '@@@OUTPUT' && opam list -s --color=never --depends-on 'a.0.0.1' --coinstallable-with 'a.0.0.1' --all-versions --depopts && opam list -s --color=never --depends-on 'a.0.0.1' --coinstallable-with 'a.0.0.1' --all-versions --recursive && opam list -s --color=never --depends-on 'a.0.0.1' --coinstallable-with 'a.0.0.1' --all-versions --with-test --depopts && echo '@@@OUTPUT'
+    RUN opam pin add -k git -y opam-ci-check git+https://github.com/ocurrent/opam-repo-ci.git#live
```
punchagan, line 1109: This step could get cached and in a Dockerfile we could use the `ADD` command to work around the caching:

```dockerfile
ADD https://api.github.com/repos/punchagan/opam-repo-ci/git/refs/heads/revdeps-exp version.json
```

but obuilder spec doesn't seem to support `ADD` and I'm not sure there's a nice way to do this pinning with commit hashes since we'd want to pin to the commit after the merge.

shonfeder: We discussed three ideas in our co-working meeting:

- Extending obuilder with `add`
- Some way to ensure that previous actions in docker container would invalidate the cache when needed
- Deploying a docker image alongside the other deployment with `opam-ci-check` and using that as the base

shonfeder:
We agreed in a sync that we'd close this issue but keep the branch alive to revive the work. We just don't have the bandwidth to land this at moment, given the low priority of this addition.

PR #404, opened 2024-12-17
opam_ci_check: Add option to use default opam root instead of creating new one

punchagan:
This PR adds a CLI option to run the opam-ci-check revdeps related commands using the default opam-root instead of creating a new one. This makes the revdeps functionality easier to use in our CI images. 

The PR also has a couple of minor preparatory commits to make it easier to use the revdeps listing command in CI.


shonfeder (approved):
Looks great! Thank you for factoring this out, makes for a super easy review :)

punchagan:
> Looks great! Thank you for factoring this out, makes for a super easy review :)

Thanks for making the time to review! And asking for it to be factored out. Happy to do anything that helps make reviews easier. :)

PR #406, opened 2024-12-17
Fetch each package into its own directory when linting

### LINKED ISSUE #405: lint broken when considering multiple packages
As the story begins:
I opened a opam-repository PR at https://github.com/ocaml/opam-repository/pull/27135
I discovered there's a lot of lint issues of the form ""Error in PPP: Your dune-project file indicates that this package requires at least dune 2.0 but your opam file only requires dune >= ZZZ. Please check which requirement is the right one, and fix the other."

Now, I thought "ok, I can fix these errors easily". I took a dive into PPP's dune-project file, and discovered it says "lang dune ZZZ". So, I wondered where the "2.0" comes from.

It turns out in opam-ci-check/bin/main.ml, you have the following code:
```OCaml
      Dir_helpers.with_temp_dir "opam-ci-check-lint-" @@ fun dir ->
      let process_package { pkg; src; newly_published } =
        let opam = read_package_opam ~opam_repo_dir pkg in
        let pkg_src_dir =
          if Option.is_none src then fetch_package_src ~dir ~pkg opam else src
        in
        Lint.v ~pkg ~newly_published ~pkg_src_dir opam
      in
      let all_lint_packages = List.map process_package package_specs in
      let errors = Lint.lint_packages ~opam_repo_dir all_lint_packages in
```

Now, we have a list of packages to lint. Ok. We create _one_ temporary directory (`dir`) where we download (fetch_package_src) the package contents to.

I guess since the same temporary directory will be used, this will overwrite all the files in that directory -- so the dune-project file only of the very last package.

So maybe you can adjust your code to go full circle with one package, or download into separate directories? Thanks a lot.

shonfeder:
Closes #405

Follows @punchagan's comment at https://github.com/ocurrent/opam-repo-ci/issues/405#issuecomment-2549278971

I've tested locally that this works on
https://github.com/ocaml/opam-repository/pull/27138

### DIFF
--- a/opam-ci-check/bin/main.ml
+++ b/opam-ci-check/bin/main.ml
@@ -72,7 +72,11 @@ let lint package_specs local_repo_dir =
       let process_package { pkg; src; newly_published } =
         let opam = read_package_opam ~opam_repo_dir pkg in
         let pkg_src_dir =
-          if Option.is_none src then fetch_package_src ~dir ~pkg opam else src
+          if Option.is_none src then
+            let dir = dir // OpamPackage.to_string pkg in
+            fetch_package_src ~dir ~pkg opam
+          else
+            src
         in
         Lint.v ~pkg ~newly_published ~pkg_src_dir opam
       in


punchagan (approved):
Thanks for taking care of this!

PR #412, opened 2025-01-01
Add lint checks for the archive repo

### LINKED ISSUE #411: Add lint check for archive repo
As per the [archiving plan](https://github.com/ocaml/opam-repository/wiki/Package-Archiving:-Plan) we need a lint check that will

- [x] checks that every dependency constraint is of the form "dep" {... & <= "<latest version of dep available in opam-repository at the time of the PR or lower>"} except for "ocaml" (needed for phase 2)
- [x] checks if the x-reason-for-archival is set (needed for phase 3)
- [x] checks if the x-opam-repository-commit-hash-at-time-of-archival is set (needed for phase 3)

in accordance with the [archiving policy](https://github.com/ocaml/opam-repository/blob/master/governance/policies/archiving.md#archiving-a-package)

We could do this in a standalone tool, but since the purpose of opam-ci-check is to support the linting requirements for the opam repo, and this is an extension of that, I think it is the right place.

Once #412 is merged, I'll just need to update the Git workflow at https://github.com/ocaml/opam-repository-archive/blob/04de824d1d8e75e729440d214181556fc613e7b6/.github/workflows/main.yml#L34 to use the `live` branch of ocurrent.

shonfeder:
Closes #411

To support using the tool in different contexts, this change also extends the
API and CLI to allow specifying groups of related lints to be run. The default
falls back to the current behavior. This may also be useful for down-streaming
the linting into publication tools.

I suggest reviewing by commit, at least initially, since the change sets are
all logically grouped. The diff is not as long as it looks, since 177 lines
are coming from the cram tests.


shonfeder:
I need to add another layer of logic here that will bypass fetching sources when we are doing the simpler set of checks, otherwise we waste a lot of http requests for no reason. I'll do that tomorrow.

shonfeder:
This has been tested in the intended CI target: https://github.com/ocaml/opam-repository-archive/pull/3

opam-ci-check/lib/lint_error.ml
```diff
@@ -37,9 +37,15 @@ type error =
   | PrefixConflictClassMismatch of prefix_conflict_class_mismatch
   | DefaultTagsPresent of string list
   | MissingUpperBound of string
+  | InvalidReasonForArchiving
```
punchagan, line 40: Would it be better to have the reason string in the error to make it easier to catch subtle typos?

opam-ci-check/test/lint.t
```diff
@@ -344,3 +363,149 @@ passes linting:
...
+  > packages/a-1/a-1.0.0.1/opam > opam.new
+  $ mv opam.new packages/a-1/a-1.0.0.1/opam
+  $ opam-ci-check lint -r . --check=archive-repo a-1.0.0.1
+  Linting opam-repository at $TESTCASE_ROOT/. ...
+  Error in a-1.0.0.1: An upper bound constraint is missing on dependency 'baz'
+  Error in a-1.0.0.1: An upper bound constraint is missing on dependency 'bar'
+  Error in a-1.0.0.1: An upper bound constraint is missing on dependency 'foo'
+  [1]
+
+Test that we do NOT report errors when all a packages dependandies have an upper bound:
```
punchagan, line 416: Typo: - s/a packages dependandies/package dependencies/

opam-ci-check/test/lint.t
```diff
@@ -344,3 +363,149 @@ passes linting:
...
+  x-opam-repository-commit-hash-at-time-of-archiving: "de786e28dbea73843ad5e5f0290a4e81fba39370"
+  $ opam-ci-check lint -r . --check=archive-repo a-1.0.0.1
+  Linting opam-repository at $TESTCASE_ROOT/. ...
+  No errors
+
+## Upper bounds checks
+
+All dependencies
+
+Test that we report errors when a package has dependandies without an upper bound:
```
punchagan, line 403: Typo: dependandies

punchagan (approved):
Thanks for adding this functionality. The changes look good to me, apart from a couple of minor comments.

opam-ci-check/lib/lint_error.ml
```diff
@@ -37,9 +37,15 @@ type error =
...
 
 (**/**)
 
+let x_reason_for_archiving_field = "x-reason-for-archiving"
+(* Used in the opam repo archive *)
+let x_reason_for_archiving_valid_reasons =
+  ["ocaml-version"; "source-unavailable"; "maintenance-intent"; "uninstallable" ]
```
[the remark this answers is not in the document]
shonfeder: Ohh, we didn't even consider that other meaning! `not-installable` is definitely clearer. This is actually down stream of the opam repo policy. It might not be worth changing now, but if you think it could cause problems opening a PR/issue into the policy is where to start.

opam-ci-check/lib/lint_error.ml
shonfeder: I did have that initially, but it added a relatively high amount of complexity to the `check_` function, and I didn't think it was worth it given this tool is currently only for use by an expert, and there are only 4 reasons to check. So, in short, I think it is not worth doing that currently, but if we add more reasons or we end up having non-experts archiving packages, we may want to add that.

opam-ci-check/test/lint.t
shonfeder, line 416:
```suggestion
Test that we do NOT report errors when all a packages dependencies have an upper bound:
```

opam-ci-check/test/lint.t
shonfeder, line 403:
```suggestion
Test that we report errors when a package has dependencies without an upper bound:
```

shonfeder:
Thank you for the review! :pray:

opam-ci-check/lib/lint_error.ml
punchagan: Fair enough!

opam-ci-check/lib/lint_error.ml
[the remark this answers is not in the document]
punchagan: Yeah, it's probably not worth changing now. And I don't think it would cause any serious problems.

PR #416, opened 2025-01-26
Update lower bound of ocaml-version

shonfeder:
This is a followup to https://github.com/ocurrent/opam-repo-ci/pull/414

Updating the lower bound should probably be part of the routine update
procedure, because once the service is deployed with updated ocaml versions, we
really do need the latest release of `ocaml-version` installed for the program
to behave correctly.


punchagan (approved):
We could add a daily/weekly cron GHA to open PRs updating the opam files if the `ocaml-version` has had a new release.

PR #417, opened 2025-01-26
Extract opam-ci-check into it's own library

shonfeder:
This builds on @punchagan's work, but makes the isolation of the lib starker, by moving it into a separate directory. The latter allows for a simpler build rule, and makes the separations of concerns clear from the directory structure.

If @punchagan is happy with this change, we should squash the changes on merge.


punchagan (approved):

punchagan:
Thanks for the PR! It does improve the clarity of separation of concerns! :tada:

PR #421, opened 2025-02-10
Don't raise a linting error on conf- packages without sources

### LINKED ISSUE #420: do not fail in CI when package source is not provided for `conf-` packages
For `conf-*` packages, the CI is always giving the following error:

> Error in conf-X: No package source directory provided.

It would be great to remove this error for `conf-*` packages (those having the `conf` flag in the opam file) in order to unclutter the output.

Ported from https://github.com/ocaml/opam-repository/issues/27404

shonfeder:
Closes #420 

The linting checks of conf- packages currently suffer from two related problems:

1. We incorrectly signal linting when package sources are missing, even tho conf- packages should (generally) not have sources.
2. There is no check to ensure that conf- packages are correctly specified, as per the opam manual.

The [opam manual stipulates](https://opam.ocaml.org/doc/Manual.html#opamflag-conf) that

> the package may not include a source URL or install anything, but just do some checks, and fail to install if they don't pass. conf packages should have a name starting with conf-, and include the appropriate [depexts:](https://opam.ocaml.org/doc/Manual.html#opamfield-depexts) field

This PR fixes (1) by adding an exception to the checks for the presence of sources, and fixes (2) by adding a check to ensure that conf- packages are specified correctly.

cc: @zapashcanon, @mseri, @raphael-proust

### DIFF
--- a/opam-ci-check/test/lint.t
+++ b/opam-ci-check/test/lint.t
@@ -509,3 +509,88 @@ has an invald value:
   Linting opam-repository at $TESTCASE_ROOT/. ...
   Error in a-1.0.0.1: The field 'x-opam-repository-commit-hash-at-time-of-archiving' must be present and hold a string recording the commit hash of the primary repo at the time the package version is archived.
   [1]
+
+## conf- package checks
+
+Reset the project repo
+
+  $ git reset -q --hard initial-state
+
+Test that a package with a conf- prefix but missing the `conf` flag and `depext`
+field triggers the expected lint errors:
+
+  $ export INVALID_CONF_DIR=packages/conf-invalid/conf-invalid.0.0.1/
+  $ mkdir -p $INVALID_CONF_DIR
+  $ cat > $INVALID_CONF_DIR/opam <<EOF
+  > opam-version: "2.0"
+  > synopsis: "Synopsis"
+  > description: "Description"
+  > maintainer: "Maintainer <me@example.com>"
+  > author: "Author"
+  > license: "Apache-2.0"
+  > homepage: "https://github.com/ocurrent/opam-repo-ci"
+  > bug-reports: "https://github.com/ocurrent/opam-repo-ci/issues"
+  > dev-repo: "git+https://github.com/ocurrent/opam-repo-ci.git"
+  > doc: "https://ocurrent.github.io/ocurrent/"
+  > build: []
+  > depends: []
+  > EOF
+  $ opam-ci-check lint -r . conf-invalid.0.0.1
+  Linting opam-repository at $TESTCASE_ROOT/. ...
+  Error in conf-invalid.0.0.1: No package source directory provided.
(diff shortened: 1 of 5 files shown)


opam-ci-check/test/lint.t
```diff
@@ -509,3 +509,88 @@ has an invald value:
...
+  > dev-repo: "git+https://github.com/ocurrent/opam-repo-ci.git"
+  > doc: "https://ocurrent.github.io/ocurrent/"
+  > build: []
+  > depends: []
+  > depexts: [ "curl" ]
+  > flags: [ conf ]
+  > EOF
+  $ opam-ci-check lint -r . conf-valid.0.0.1
+  Linting opam-repository at $TESTCASE_ROOT/. ...
+  No errors
```
shonfeder, line 595: @mseri, @raphael-proust if either of you could review the linting errors here to confirm that you are happy with this kind of feedback, from the perspective of opam repo maintainers, that would be helpful confirmation that we're doing the right thing here.

Thanks!

opam-ci-check/test/lint.t
```diff
@@ -509,3 +509,88 @@ has an invald value:
...
+  > homepage: "https://github.com/ocurrent/opam-repo-ci"
+  > bug-reports: "https://github.com/ocurrent/opam-repo-ci/issues"
+  > dev-repo: "git+https://github.com/ocurrent/opam-repo-ci.git"
+  > doc: "https://ocurrent.github.io/ocurrent/"
+  > build: []
+  > depends: []
+  > EOF
+  $ opam-ci-check lint -r . conf-invalid.0.0.1
+  Linting opam-repository at $TESTCASE_ROOT/. ...
+  Error in conf-invalid.0.0.1: No package source directory provided.
```
mseri, line 540:
```suggestion
  Error in conf-invalid.0.0.1: No package source directory provided.
```
This is marked error here because the conf flag is missing, right?

opam-ci-check/test/lint.t
```diff
@@ -509,3 +509,88 @@ has an invald value:
...
+  > bug-reports: "https://github.com/ocurrent/opam-repo-ci/issues"
+  > dev-repo: "git+https://github.com/ocurrent/opam-repo-ci.git"
+  > doc: "https://ocurrent.github.io/ocurrent/"
+  > build: []
+  > depends: []
+  > depexts: [ "curl" ]
+  > EOF
+  $ opam-ci-check lint -r . conf-invalid.0.0.2
+  Linting opam-repository at $TESTCASE_ROOT/. ...
+  Error in conf-invalid.0.0.2: No package source directory provided.
```
mseri, line 567: This is marked error here because the conf flag is missing, right?

opam-ci-check/test/lint.t
mseri: If my understanding above is correct, then I think this is good

opam-ci-check/test/lint.t
shonfeder: Correct!

mseri (approved):

mtelvers (approved):

shonfeder:
Thank you for the reviews!

PR #433, opened 2025-05-23
Remove DuneLowerBoundMissing linter variant

### LINKED ISSUE #432: The `dune` dependency bound check gives false positives
See https://github.com/ocaml/opam-repository/pull/27899#discussion_r2100430302

I wonder if we can just get rid of this check in general now, since, IIUC, it should already be accounted for by our lower-bounds check in case someone does not specify a lower bound.

ivnhk:
Fixes #432

Try to remove the linter check for `dune` lower bound. We will try to handle this by utilizing more general lower bounds check.


ivnhk:
For some reason, `ocaml-CI` keeps failing and I can't quite figure out what is the reason

ivnhk:
Unfortunately, some builds are still failing, namely:
**- EXPERIMENTAL (lower-bound)
- MacOS (4 builds)
- EXPERIMENTAL openbsd builds (2 builds)
- EXPERIMENTAL windows-server (2 builds)**

Idk whether it should be like this

opam-ci-check/lint/opam_ci_check_lint.ml
```diff
@@ -271,13 +271,13 @@ module Checks = struct
...
         | None, Ok None -> []
-        | Some "", _ -> [ (pkg, DuneLowerBoundMissing) ]
         | Some _, Ok None -> [ (pkg, DuneProjectMissing) ]
         | None, Ok (Some _) ->
           if is_dune (OpamPackage.name pkg) then []
           else [ (pkg, DuneDependencyMissing) ]
         | Some dep, Ok (Some ver) ->
-          if OpamVersionCompare.compare dep ver >= 0 then []
+          if dep = "" && ver = "1.0" then []
+          else if OpamVersionCompare.compare dep ver >= 0 then []
```
shonfeder, line 280: This is a nit, so feel free to disregard --

While the logic here is pretty pretty straight forward, I have a nagging feeling it may not be as clear as we could get. Here's an idea to try to make the meaning here more apparent:

When `dep = ""`, we take this to mean that its dependency is a lower bound on `>= 1.0`. That is captured implicitly on line 279. We could make this explicit by adding a comment on this line, but I think we can make this explicit in the code, while also reducing the number of branches in the conditional, via

```suggestion
          let dune_dep_lower_bound = if dep = "" then "1.0" else dep in
          if OpamVersionCompare.compare dune_dep_lower_bound ver >= 0 then []
```

shonfeder (commented):
Just two suggestions here. Only the first needs to be addressed, and in both cases I'm open to alternative if you see a better way!

shonfeder (approved):
Looks great. Awesome work here and thank you for the fix!!

shonfeder:
The CI failures here are not related to this changeset.

Issue #400, opened 2024-12-11
opam-ci-check is surprisingly slow when checking several packages


shonfeder:
Using the linter on https://github.com/dra27/opam-repository/pull/20, which as 47 packages, took about 1 minute. This surprised me, as I don't think the linting is doing anything computationally expensive.

We likely have some room for low-hanging perf improvements. Perhaps this would start with making the executable run tasks concurrently, but some benchmarking should guide us.

This is low-priority, and not in scope for this quarter, IMO, since it is not a problem for 99% of package publication use cases, and never going to be bottleneck on the CI for publication of large packages. I'm just opening this to record the finding :)

punchagan:
> Using the linter on [dra27/opam-repository#20](https://github.com/dra27/opam-repository/pull/20), which has 47 packages, took about 1 minute. This surprised me, as I don't think the linting is doing anything computationally expensive.

Is a bulk of the time being spent in downloading the sources of the packages? (for the dune checks?) Doing that concurrently could help...

punchagan:
I was testing for #405 and ran into this myself too. 

I didn't do any detailed benchmarking but from crude printf benchmarking, it looks like we spend a bunch of time downloading the package sources and in running [`opam-lint` with `--check-upstream`](https://opam.ocaml.org/doc/man/opam-lint.html) which checks if the upstream archive is available and the checksum matches.  Like you said, running these tasks concurrently should easily give us some improvements.

PR #442, opened 2025-06-19
Remove double checks for source archive checksums

punchagan:
The [opam lint with `--check-upstream`](https://github.com/ocurrent/opam-repo-ci/blob/12810c8cc0b9a02e7b22f660387b76c6e70aa685/opam-ci-check/lint/opam_ci_check_lint.ml#L313) ensures that the [checksums for the upstream archive](https://github.com/ocaml/opam/blob/39165180cda81e8778b2ac9bb6eab363eac92261/src/state/opamFileTools.ml#L823-L850) is correct. But, we also have a separate `check_checksums` that also ensures this. This commit removes the double checksum check in the default case, but keeps the check when running with only "--checks opam-file".

This partially addresses #400.

### DIFF
--- a/opam-ci-check/bin/main.ml
+++ b/opam-ci-check/bin/main.ml
@@ -338,7 +338,7 @@ let lint_cmd =
         "primary-repo", Opam_repo_publication;
         "archive-repo", Opam_repo_archive])
     in
-    let defaults = Lint.Checks.[General_opam_file; Opam_repo_publication] in
+    let defaults = Lint.Checks.[Opam_repo_publication] in
     Arg.value (Arg.opt options defaults info)
   in
   let term =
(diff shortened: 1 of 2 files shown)


punchagan:
I hadn't looked carefully at what the `check_checksums` does and was working from my memory. We don't actually run the checksum checks in that check -- it just checks that we have stronger checksums than md5. So, this PR is incorrect, and I'm closing this. 

But, we also do run an additional checksum check when downloading the sources in [`fetch_package_src`](https://github.com/ocurrent/opam-repo-ci/blob/12810c8cc0b9a02e7b22f660387b76c6e70aa685/opam-ci-check/bin/main.ml#L27). So, there's possibly work that can be done to speed up things a little.

PR #444, opened 2025-06-26
lower-bounds: run the check with the opam-repository-archive on

mseri:
Currently the opam-repository will test the lower bound on most latest versions of packages. And even though most cases are avoided by the fact that we add upper bounds to all archived packages, the new packages may still be compatible and co-installable. So we should keep checking the lower bounds with the whole database to ensure that the metadata on the opam-repository remains consistent.

This will also help to ensure that all the people that use the archive in their CI do not have unpleasant surprises.


shonfeder (approved):

shonfeder:
Thanks! I'm not 100% convinced that we want to keep testing stuff form the archive long term, but I also don't see any harm in adding this for the moment.

PR #461, opened 2025-10-30
Update docs and tests with latest ocaml-version and dockerfile

punchagan:


punchagan:
I've included the updates in #457

PR #482, opened 2026-07-11
index: use GLOB not LIKE in get_full_hash to hit the PK index

mtelvers:
`get_full_hash` ran:

```sql
SELECT DISTINCT hash FROM ci_build_index WHERE owner = ? AND name = ? AND hash LIKE ?
```

SQLite's `LIKE` is **case-insensitive by default**, which disables the index for the `hash` term. So the query seeks to the `owner/name` prefix and then **scans every row under it** — for `ocaml/opam-repository` that is the entire **4.5M-row** `ci_build_index`. Because `get_full_hash` runs on every commit/variant page and the SQLite call is synchronous on the single Lwt thread, ~1 req/s saturates a core (measured **~0.97s CPU per call**), so any crawl of the web UI makes the engine unresponsive.

`GLOB` is case-sensitive, so its prefix pattern compiles to an **indexed range seek** (`hash>? AND hash<?`). Git hashes are lowercase hex, so case-sensitivity is a non-issue and the prefix / `Ambiguous` semantics are unchanged.

**Measured on the production 6.2 GB index** (`EXPLAIN QUERY PLAN` + `.timer`):

| query | plan | time |
|---|---|---|
| `hash LIKE 'H%'` (before) | `SEARCH (owner=? AND name=?)` → scan | 968 ms |
| `hash GLOB 'H*'` (after) | `SEARCH (… AND hash>? AND hash<?)` → range seek | 1 ms |

~900× faster; identical for full and short hashes. Root-caused via `perf` (96.9% of engine CPU was in libsqlite3, dominated by `VdbeExec`/`RecordCompareWithSkip`/`BtreeNext`).

### DIFF
--- a/lib/index.ml
+++ b/lib/index.ml
@@ -59,8 +59,13 @@ CREATE TABLE IF NOT EXISTS ci_build_index (
                                      WHERE owner = ? AND name = ? AND hash = ? AND variant = ?" in
   let get_job_ids = Sqlite3.prepare db "SELECT variant, job_id FROM ci_build_index \
                                      WHERE owner = ? AND name = ? AND hash = ?" in
+  (* GLOB (not LIKE) so the query uses the primary-key index: LIKE is
+     case-insensitive by default, which disables the index and forces a full
+     scan of the (millions-of-rows) table on every call. GLOB is case-sensitive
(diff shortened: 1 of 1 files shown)


punchagan (approved):
Nice catch! Thanks for the fix, @mtelvers !

PR #483, opened 2026-07-25
Retry opam init to survive the opam 2.6 root layout upgrade

mtelvers:
`opam-dev` in the base images is now `2.6.0~alpha1`, which bumps the opam root layout from the version 2.1 the images ship with. On the first invocation `opam init --reinit` performs the migration, prints *"Update done, please now retry your command"* and exits **10** (Aborted), so every job using opam dev fails at setup — which, since `lib/build.ml` sets `opam_version = ` `` `Dev `` globally, is every build, test, lower-bounds and revdep job:

```
/home/opam: (run (network host)
                 (shell "opam init --reinit --config .opamrc-sandbox -ni"))
...
This version of opam requires an update to the layout of /home/opam/.opam from version 2.1 to version 2.6~alpha, which can't be reverted.
Perform the update and continue? [Y/n] y
...
Upgrading the internal repository format for 'default'...
Format upgrade done.

<><> Rerunning init and update ><><><><><><><><><><><><><><><><><><><><><><><><>
...
Update done, please now retry your command.
"/usr/bin/env" "bash" "-c" "opam init --reinit --config .opamrc-sandbox -ni" failed with exit status 10
```

Running the identical command a second time returns 0, and no flag suppresses the abort. Verified in `ocaml/opam:debian-testing-ocaml-4.14@sha256:00036ba6…`:

| attempt | exit |
| --- | --- |
| `opam init --reinit --config .opamrc-sandbox -ni` | 10 |
| … `-y` | 10 |
| … `--confirm-level=unsafe-yes` | 10 |
| … `OPAMNOAUTOUPGRADE=1` | 10 |
| … `--cli=2.1` | 10 |
| … `\|\| opam init --reinit … -ni` | **0** |

`--no-auto-upgrade`/`OPAMNOAUTOUPGRADE` only governs the *repository* format, not the root, and `opam init --help` has nothing for the root upgrade. Exit 10 is documented as *"Aborted. The operation required confirmation, which wasn't given"*, which is misleading here: the confirmation was auto-answered `y` and the migration succeeded. Arguably an opam bug — for `init` the command you are told to retry is the one that just ran — so this is a stopgap, and the commit message says to revert it once opam stops aborting. I'll raise it upstream.

The retry is conditional (`||`) rather than an unconditional second run because a successful `opam init --reinit` already updates the repositories, which fetches `ocaml-patches-overlay` from GitHub; running it twice on the opam versions needing no migration would double that traffic on every job, which is what `OPAMDOWNLOADJOBS=1` exists to avoid.

Released opam is unaffected — I checked `opam-2.5` (2.5.2) inits cleanly first time — so the `` `V2_1 ``…`` `V2_5 `` matrix jobs never hit the second invocation. After the fix the following `opam option solver=builtin-0install` step succeeds and the base packages survive the migration (`base-bigarray base-threads base-unix ocaml ocaml-base-compiler ocaml-config ocaml-options-vanilla opam-depext`).

Note this does **not** invalidate any cached results: `Cluster_build.Op.Key` is `{pool; commit; variant; ty}` and the spec string reaches `Op.build` through the unhashed `Op.t` context, so changing the spec text does not re-run existing jobs. (Same reason the weekly base-image digest refresh doesn't.) Jobs that have already failed with exit 10 keep their recorded failure until they are pushed to or rebuilt.

Tests promoted: `opam-ci-check/test/build.t` and `test/specs.expected`.

### DIFF
--- a/opam-ci-check/lib/opam_build.ml
+++ b/opam-ci-check/lib/opam_build.ml
@@ -108,7 +108,14 @@ let setup_repository ?(local=false) ~variant ~for_docker ~opam_version () =
   (* TODO: MacOS seems to have a bug in (copy ...) so I am forced to remove the (workdir ...) here.
      Otherwise the "opam pin" after the "opam repository set-url" will fail (cannot find the new package for some reason) *)
   run "%s -f %s/bin/opam-%s %s/bin/opam" ln prefix opam_version_str prefix ::
-  run ~network "opam init --reinit%s -ni" opamrc :: (* TODO: Remove ~network when https://github.com/ocurrent/ocaml-dockerfile/pull/132 is merged *)
+  (* NOTE: opam 2.6~alpha bumps the opam root layout from the 2.1 version the
+     base images ship with. On the first run it performs the migration and then
+     exits 10 (Aborted) with "Update done, please now retry your command", so we
+     have to run it again. No flag suppresses this (-y, --confirm-level,
+     OPAMNOAUTOUPGRADE and --cli=2.1 all still abort). The retry is conditional
+     so that opam versions needing no migration don't update the repositories,
+     and hit GitHub, twice per job. *)
+  run ~network "opam init --reinit%s -ni || opam init --reinit%s -ni" opamrc opamrc :: (* TODO: Remove ~network when https://github.com/ocurrent/ocaml-dockerfile/pull/132 is merged *)
   run "opam option solver=builtin-0install && opam config report" ::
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
2.6.0~beta1 has fixed this so this shouldn't be necessary once the docker-base-images have been rebuilt and should be removed from `live` as well.

mtelvers:
Closing as no longer needed.

